PR #2, opened 2019-01-01
Implementing the second-order equations for calculating pressure and temperature.

Matt4890:
### Warning:
This code has not been tested as of yet on a flight board. Also, for some reason, my error checking isn't working, so beware minor syntax errors.

# Changes Made
1. Implemented second-order equations for pressure and temperature
    - Compensation for non-linearity at low (< 20°C) and very low (< -15°C) temperatures as specified by the data sheet
2. Formatted and commented the entire file
    - Indentation
    - Sections
    - Function descriptions
    - Etc.
3. Simple optimizations
    - Replaced multiplication or division by pow(2, x) with logical shifts
        - One assmebly instruction, as opposed to jumping to a generic library instruction
        - Allowed the replacement of double types with int32_t types
            - Less casting
            - Fewer overall variables
    - When receiving data from the barometer, `var = 0; var += data` has been replaced with `var = data`
4. Fixed confusing semantics
    - After sending the ADC or PROM read command, the ADC read command was repetedly being sent to obtain each byte. While this technically works, the semantics became very confusing when reading from PROM. After starting a read command, 0x00 should be sent in order to recieve a byte from the barometer. It just so happens that the command to start and ADC read was also 0x00. Instances of this have now been replaced with a more descriptive `READ_BYTE_CMD` constant, which also has the value 0x00.

### DIFF
--- a/Src/ReadBarometer.c
+++ b/Src/ReadBarometer.c
@@ -1,11 +1,13 @@
 /**
- * This file contains constants and functions designed to obtain accurate pressure
- * and temperature readings from the MS5607-02BA03 barometer on the board.
- */
+  ******************************************************************************
+  * File Name          : ReadBarometer.c
+  * Description        : This file contains constants and functions designed to
+  *                      obtain accurate pressure and temperature readings from
+  *                      the MS5607-02BA03 barometer on the board.
+  ******************************************************************************
+*/
 
-//
-// Global Headers
-//
+/* Includes ------------------------------------------------------------------*/
 
 #include "stm32f4xx.h"
 #include "stm32f4xx_hal_conf.h"
@@ -15,22 +17,15 @@
 #include "ReadBarometer.h"
 #include "Data.h"
 
-//
-// Local Headers
-//
+/* Macros --------------------------------------------------------------------*/
 
-uint16_t readCalibrationCoefficient(const uint8_t PROM_READ_CMD);
+#define READ_BAROMETER_PERIOD   20  // Sampling delay set to 50 Hz to match high frequency logging
+#define TEMP_VERY_LOW           -1500
+#define TEMP_LOW                2000
+#define CMD_SIZE                1
+#define CMD_TIMEOUT             150
 
-//
-// Constants
-//
-
-static const int READ_BAROMETER_PERIOD      = 20;    // Sampling delay set to 50 Hz to match high frequency logging
-static const int TEMP_VERY_LOW              = -1500;
-static const int TEMP_LOW                   = 2000;
-
-static const int CMD_SIZE                   = 1;
-static const int CMD_TIMEOUT                = 150;
+/* Constants -----------------------------------------------------------------*/
 
 static const uint8_t ADC_D1_512_CONV_CMD    = 0x42;
 static const uint8_t ADC_D2_512_CONV_CMD    = 0x52;
@@ -43,6 +38,16 @@ static const uint8_t PROM_READ_TREF_CMD     = 0xAA;
 static const uint8_t PROM_READ_TEMPSENS_CMD = 0xAC;
 static const uint8_t READ_BYTE_CMD          = 0x00;
 static const uint8_t RESET_CMD              = 0x1E;
+
+/* Variables -----------------------------------------------------------------*/
+
+/* Structs -------------------------------------------------------------------*/
+
+/* Prototypes ----------------------------------------------------------------*/
+
+uint16_t readCalibrationCoefficient(const uint8_t PROM_READ_CMD);
+
+/* Functions -----------------------------------------------------------------*/
 
 /**
  * This function is to be used as a thread task that reads and updates
@@ -108,16 +113,14 @@ void readBarometerTask(void const* arg)
         // Delay so that the loop operates at 50Hz
         osDelayUntil(&prevWakeTime, READ_BAROMETER_PERIOD);
 
-        //
-        // Read Digital Pressure (D1)
-        //
+        /* Read Digital Pressure (D1) ----------------------------------------*/
 
-        // Tell the barometer to convert the pressure to a digital value with an OSR of 512
+        // Tell the barometer to convert the pressure to a digital value with an over-sampling ratio of 512
         HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_RESET);
         HAL_SPI_Transmit(&hspi2, &ADC_D1_512_CONV_CMD, CMD_SIZE, CMD_TIMEOUT);
         HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_SET);
 
-        osDelay(2); // 1.17ms max conversion time for OSR 512
+        osDelay(2); // 1.17ms max conversion time for an over-sampling ratio of 512
 
         // Read the pressure value
 
@@ -142,16 +145,14 @@ void readBarometerTask(void const* arg)
         // Release chip select to end the command
         HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_SET);
 
-        //
-        // Read Digital Temperature (D2)
-        //
+        /* Read Digital Temperature (D2) -------------------------------------*/
 
-        // Tell the barometer to convert the temperature to a digital value with an OSR of 512
+        // Tell the barometer to convert the temperature to a digital value with an over-sampling ratio of 512
         HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_RESET);
         HAL_SPI_Transmit(&hspi2, &ADC_D2_512_CONV_CMD, CMD_SIZE, CMD_TIMEOUT);
         HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_SET);
 
-        osDelay(2); // 1.17ms max conversion time for OSR 512
+        osDelay(2); // 1.17ms max conversion time for an over-sampling ratio of 512
 
         // Read the temperature value
 
@@ -176,9 +177,7 @@ void readBarometerTask(void const* arg)
         // Release chip select to end the command
         HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_SET);
 
-        //
-        // Calculate First-Order Temperature and Pressure
-        //
+        /* Calculate First-Order Temperature and Pressure --------------------*/
 
         // Equations provided by MS5607-02BA03 data sheet
 
@@ -188,9 +187,7 @@ void readBarometerTask(void const* arg)
         int64_t sens    = (c1Sens << 16) + ((dT * c3Tcs) >> 7);
         int32_t p       = (((pressureReading * sens) >> 21) - off) >> 15;   // Divide this value by 100 to get millibars
 
-        //
-        // Calculate Second-Order Temperature and Pressure
-        //
+        /* Calculate Second-Order Temperature and Pressure -------------------*/
 
         // Equations provided by MS5607-02BA03 data sheet
 
@@ -211,23 +208,18 @@ void readBarometerTask(void const* arg)
             sens    = sens  - sens2;
         }
 
-        //
-        // Store Data
-        //
+        /* Store Data --------------------------------------------------------*/
 
-        // Check if the mutex is available
-        if (osMutexWait(data->mutex_, 0) != osOK)
+        if (osMutexWait(data->mutex_, 0) == osOK)
         {
-            // Loop and get new values if the mutex was taken
-            continue;
+            // Assign new values if mutex was available
+            data->pressure_     = p;
+            data->temperature_  = temp;
+            osMutexRelease(data->mutex_);
         }
 
-        // Assign new values if mutex was available
-        data->pressure_     = p;
-        data->temperature_  = temp;
-        osMutexRelease(data->mutex_);
-
-        // PRESSURE AND TEMPERATURE VALUES ARE STORED AT 100x VALUE TO MAINTAIN 2 DECIMAL POINTS OF PRECISION AS AN INTEGER
+        // PRESSURE AND TEMPERATURE VALUES ARE STORED AT 100x VALUE TO MAINTAIN
+        // 2 DECIMAL POINTS OF PRECISION AS AN INTEGER!
         // E.x. The value 1234 should be interpreted as 12.34
     }
 }
@@ -244,9 +236,7 @@ uint16_t readCalibrationCoefficient(const uint8_t PROM_READ_CMD)
     uint16_t    coefficient;
     uint8_t     dataInBuffer;
 
-    //
-    // Read Coefficient
-    //
+    /* Read Coefficient ------------------------------------------------------*/
 
     // Chip select to start a new command
     HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_RESET);
--- a/Src/ReadBarometer.c
+++ b/Src/ReadBarometer.c
@@ -19,13 +19,13 @@
 
 /* Macros --------------------------------------------------------------------*/
 
-#define READ_BAROMETER_PERIOD   20  // Sampling delay set to 50 Hz to match high frequency logging
-#define TEMP_VERY_LOW           -1500
-#define TEMP_LOW                2000
-#define CMD_SIZE                1
-#define CMD_TIMEOUT             150
-
 /* Constants -----------------------------------------------------------------*/
+
+static const int READ_BAROMETER_PERIOD      = 20;   // Sampling delay set to 50 Hz to match high frequency logging
+static const int TEMP_LOW                   = 2000;
+static const int TEMP_VERY_LOW              = -1500;
+static const int CMD_SIZE                   = 1;
+static const int CMD_TIMEOUT                = 150;
 
 static const uint8_t ADC_D1_512_CONV_CMD    = 0x42;
 static const uint8_t ADC_D2_512_CONV_CMD    = 0x52;
@@ -108,7 +108,7 @@ void readBarometerTask(void const* arg)
      *
      * Restricted to 50Hz.
      */
-    while (1)
+    for (;;)
     {
         // Delay so that the loop operates at 50Hz
         osDelayUntil(&prevWakeTime, READ_BAROMETER_PERIOD);
--- a/Src/ReadBarometer.c
+++ b/Src/ReadBarometer.c
@@ -25,24 +25,24 @@ uint16_t readCalibrationCoefficient(const uint8_t PROM_READ_CMD);
 // Constants (Could these be replaced with #define?)
 //
 
-static const int READ_BAROMETER_PERIOD		= 20;    // Sampling delay set to 50 Hz to match high frequency logging
-static const int TEMP_VERY_LOW				= -1500;
-static const int TEMP_LOW					= 2000;
+static const int READ_BAROMETER_PERIOD      = 20;    // Sampling delay set to 50 Hz to match high frequency logging
+static const int TEMP_VERY_LOW              = -1500;
+static const int TEMP_LOW                   = 2000;
 
-static const int CMD_SIZE					= 1;
-static const int CMD_TIMEOUT				= 150;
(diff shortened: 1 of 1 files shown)


Src/ReadBarometer.c
```diff
@@ -6,127 +15,256 @@
...
+		//
+		// Calculate First-Order Temperature and Pressure
+		//
 
-        HAL_SPI_TransmitReceive(&hspi2, &ADC_READ_CMD, &dataInBuffer, CMD_SIZE, CMD_TIMEOUT);
-        temperatureReading += dataInBuffer;
+		// Equations provided by MS5607-02BA03 data sheet
 
-        HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_SET);
+		int32_t dT		= temperatureReading - (c5Tref << 8);
```
NataliaP721: If we do not use doubles, we are rounding intermediate values that are used in later calculations. I'm not sure what effect it will have on the accuracy of the calculations.

Src/ReadBarometer.c
```diff
@@ -6,127 +15,256 @@
...
+		int32_t temp	= 2000 + ((dT * c6Tempsens) >> 23);					// Divide this value by 100 to get degrees Celcius
+		int32_t off		= (c2Off << 17) + ((dT * c4Tco) >> 6);
+		int32_t sens	= (c1Sens << 16) + ((dT * c3Tcs) >> 7);
+		int32_t p		= ((pressureReading * sens) >> 21) - off) >> 15;	// Divide this value by 100 to get millibars
```
NataliaPavlovic: Needs another '(' before pressureReading

Src/ReadBarometer.c
NataliaPavlovic: If we do not use doubles, we are rounding intermediate values that are used in later calculations. This can reduce the accuracy of the calculations, but I am not sure if it will cause a problem.

Matt4890: The data sheet recommends that these variables are integers (page 8). However, I also do not know how much of an issue this would cause. It also recommends that OFF and SENS are 64-bit signed integers, which I have only just noticed.

NataliaPavlovic (changes requested):
I think you should convert all tabs to 4 spaces. That is the convention we used last year (all the files already use it) and the formatting looks better on github.
I have added Enoch as a reviewer. We need to create a code style guide.

Src/ReadBarometer.c
NataliaPavlovic: Okay I missed that in the datasheet. In that case, it makes sense to use signed integers just like the datasheet does.

Matt4890:
> I think you should convert all tabs to 4 spaces. That is the convention we used last year (all the files already use it) and the formatting looks better on github. 

Done. Force push was because I made a mistake in the commit message and just changed it locally :stuck_out_tongue:

Src/ReadBarometer.c
```diff
@@ -6,127 +17,244 @@
...
+static const uint8_t ADC_READ_CMD           = 0x00;
+static const uint8_t PROM_READ_SENS_CMD     = 0xA2;
+static const uint8_t PROM_READ_OFF_CMD      = 0xA4;
+static const uint8_t PROM_READ_TCS_CMD      = 0xA6;
+static const uint8_t PROM_READ_TCO_CMD      = 0xA8;
+static const uint8_t PROM_READ_TREF_CMD     = 0xAA;
 static const uint8_t PROM_READ_TEMPSENS_CMD = 0xAC;
-static const uint8_t RESET_CMD = 0x1E;
+static const uint8_t READ_BYTE_CMD          = 0x00;
+static const uint8_t RESET_CMD              = 0x1E;
```
ExtraGravity, line 42: I would actually advise against formatting like this, it looks nice, but it isn't very practical for maintainability purposes. Every time someone adds something, they can't just write it as normal, they have to match the formatting of the other lines. And when someone adds a line that is longer than the rest, then they have to change the rest of the lines, which is not only annoying but also muddies a diff.

Src/ReadBarometer.c
```diff
@@ -6,127 +17,244 @@
 #include "ReadBarometer.h"
 #include "Data.h"
 
-static const int READ_BAROMETER_PERIOD = 20;    // Sampling delay set to 50 Hz to match high frequency logging
+/* Macros --------------------------------------------------------------------*/
+
+#define READ_BAROMETER_PERIOD   20  // Sampling delay set to 50 Hz to match high frequency logging
+#define TEMP_VERY_LOW           -1500
+#define TEMP_LOW                2000
+#define CMD_SIZE                1
```
ExtraGravity: The reason that these were `const` before and not `#define`s is because `const` is safer. `const` respects scope and is type-safe, you can read more here https://stackoverflow.com/questions/1674032/static-const-vs-define-vs-enum. 

`defines` were used in some places in the code because they were used to calculate other constants, a constant can't be calculated with a constant so a `define` is necessary (unless there's some way around this that I don't know).

Src/ReadBarometer.c
```diff
@@ -6,127 +17,244 @@
...
+     *          dT = D2 - Tref = D2 - (C5 * 2^8)
+     * TEMP - Actual temperature (-40...85°C with 0.01°C resolution)
+     *          TEMP = 20°C + (dT * TEMPSENS) = 2000 + (dT * C6)/2^23
+     * OFF  - Offset at actual temperature
+     *          OFF = OFFt1 + (TCO * dT) = (C2 * 2^17) + (C4 * dT)/2^6
+     * SENS - Sensitivity at actual temperature
+     *          SENS = SENSt1 + (TCS * dT) = (C1 * 2^16) + (C3 * dT)/2^7
+     * P    - Temperature compensated pressure (10...1200mbar with 0.01mbar resolution)
+     *          P = (D1 * SENS) - OFF = ((D1 * SENS)/2^21 - OFF)/2^15
+     */
```
ExtraGravity, line 85: I really dig this section of comments, saves someone from pulling out the datasheet which is necessary to understand this code.

Src/ReadBarometer.c
```diff
@@ -6,127 +17,244 @@
...
-        osDelay(2); // 1.17ms max conversion time for OSR 512
+        osDelay(2); // 1.17ms max conversion time for an over-sampling ratio of 512
+
+        // Read the temperature value
 
+        // Chip select to start a new command
         HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_RESET);
-        HAL_SPI_Transmit(&hspi2, &ADC_READ_CMD, CMD_SIZE, CMD_TIMEOUT);
 
-        temperatureReading = 0;
```
ExtraGravity, line 87: I think an explicit zeroing is better, it's less error-prone since we're not relying on a side effect of `temperatureReading = dataInBuffer << 16;`. Take for example if this code is changed to read more than 24 bits, and someone just adds `temperatureReading = dataInBuffer << 24;`, it'll result in quite a strange bug.

Src/ReadBarometer.c
```diff
@@ -6,127 +17,244 @@
...
+        /* Store Data --------------------------------------------------------*/
 
-        if (osMutexWait(data->mutex_, 0) != osOK)
+        if (osMutexWait(data->mutex_, 0) == osOK)
         {
-            continue;
+            // Assign new values if mutex was available
+            data->pressure_     = p;
+            data->temperature_  = temp;
+            osMutexRelease(data->mutex_);
```
ExtraGravity, line 205: Since this is a really crucial piece of code and would be very hard to debug if this was somehow not called, I would leave it outside this `if` statement.

Src/ReadBarometer.c
```diff
@@ -1,3 +1,14 @@
+/**
+  ******************************************************************************
+  * File Name          : ReadBarometer.c
+  * Description        : This file contains constants and functions designed to
```
ExtraGravity, line 4: This description makes it sound like this file only contains helper functions for this purpose, but I think it's pretty important to note that this file also contains the function that will be run in a loop for this purpose.

ExtraGravity (commented):
Overall I like the changes made, the comments have added a lot of clarity in some places, but I think some of the comments are actually harmful to the longevity of the code. A lack of comments where they would be helpful is bad, but an incorrect comment is much worse. The "correctness" of comments can't be checked by any automated process so it's really important to only comment what is necessary so it's easier to maintain. In terms of what's important, a general rule of thumb that can be followed for comments is to only comment on "why" not "what", because the code itself should be written clearly enough to explain "what". Code is also harder to read when the volume is longer than it needs to be.

For "organizational" comments, like `/* Constants -------*/`, I'm usually not a fan as it's usually self-explanatory and it becomes annoying to maintain later (like remembering to add a header comment for a new section), but I think that's more up to the conventions of the organization. 

I would say these code changes have some comments that are not necessary. but instead of pointing them all out myself I'll challenge you to take a second look and filter them out yourself.

Src/ReadBarometer.c
NataliaPavlovic: I wanted to try adding this kind of formatting this year. I like the way it organizes code and helps us see if there are any better ways to do it.

Src/ReadBarometer.c
NataliaPavlovic: I think this section does make it easier to understand the code without using the datasheet. My only concern with this is if there is a mistake we do not notice, and then people reference the mistake instead of the datasheet. Maybe this is not a valid concern because we are reviewing it anyway.

Src/ReadBarometer.c
NataliaPavlovic: I am just wondering what scenario could lead this if statement not to be called?

Src/ReadBarometer.c
ExtraGravity: I think that's a valid concern, I mean I didn't read the datasheet haha.

Src/ReadBarometer.c
ExtraGravity: Mmm, for an example, consider we decide that if the pressure spikes, we want to ignore it, someone might write.

```C
        if (osMutexWait(data->mutex_, 0) == osOK && pressure < 1000)
        {
            // Assign new values if mutex was available
            data->pressure_     = p;
            data->temperature_  = temp;
            osMutexRelease(data->mutex_);
        }
```

ExtraGravity: Actualy y'know, taking a second look at this, I think I made a comment on this more because I was more a fan of how it was written before? Less nested stuff is easier to read, and errors should be treated like special cases, I like to keep what "usually" happens in the regular flow of code. All in all, I guess commenting on this is pretty trivial anyways so it's not really important to change this.

Src/ReadBarometer.c
Matt4890: Alright. These have been changed back to `static const int`s. I will keep that in mind moving forwards.

Src/ReadBarometer.c
Matt4890: I've triple-checked that the information in this comment is correct. Hopefully it will make for a lot less window switching :stuck_out_tongue:

Src/ReadBarometer.c
Matt4890: I'm not quite sure I understand the bug that you'd be referencing.

If `temperatureReading = dataInBuffer << 24;` is added, `temperatureReading` is a `uint32_t`, meaning that the byte would be placed as normal, with each of the trailing digits being `0`. You would have to change `temperatureReading = dataInBuffer << 16;` to `temperatureReading += dataInBuffer << 16;`, but I don't see how explicit zeroing would aid this.

Or I could just be completely missing the point, which is always a possibility :stuck_out_tongue:

Src/ReadBarometer.c
Matt4890: I know that using `continue` can be confusing to newer programmers, and it's apparently bad practice (although I am a quite a big fan and use of it regularly... oops). But this just stood out to me as something that just flowed better when reading throught the code. It also removes an operation.

As for if the pressure spikes, I would then write the code as such:
```c
if (pressure < 1000 && osMutexWait(data->mutex_, 0) == osOK)
{
    data->pressure_     = p;
    data->temperature_  = temp;
    osMutexRelease(data->mutex_);
}
````
Putting the `pressure < 1000` condition first means that (at least in Java, I'm not 100% sure in C) if `pressure < 1000` fails, then `osMutexWait(data->mutex_, 0) == osOK` is never called, as the whole condition will be false anyways, meaning the mutex is never signed out. If C doesn't work like this, then I would just nest the entire if block into another where its condition is solely `if (pressure < 1000)`.

File: Src/ReadBarometer.c (hunk 8 above)
Comment (Matt4890) [reply Src/ReadBarometer.c:4]: This will be updated to be more descriptive.

File: Src/ReadBarometer.c (hunk 5 above)
Comment (NataliaPavlovic) [reply Src/ReadBarometer.c:85]: Let's keep this comment in the code. I thought about it, and I think it will save us a lot of time in the future. I checked it over, and it matches the info on the datasheet.
I think adding comments like this to other files would be a good idea too.

File: Src/ReadBarometer.c (hunk 7 above)
Comment (NataliaPavlovic) [reply Src/ReadBarometer.c:205]: Why would we ignore pressure spikes? The code was not doing that before.

File: Src/ReadBarometer.c [hunk 9]
```diff
@@ -6,127 +17,244 @@
...
-    uint32_t pressureReading = 0;   // Stores a 24 bit value
-    uint32_t temperatureReading = 0;   // Stores a 24 bit value
-
-    for (;;)
+    /**
+     * Repeatedly read digital pressure and temperature.
+     * Convert these values into their calibrated counterparts.
+     * Finally, update the data globally if the mutex is available.
+     *
+     * Restricted to 50Hz.
```
NataliaPavlovic: Frequency comment is unnecessary because it is already commented when READ_BAROMETER_PERIOD is declared.

Src/ReadBarometer.c
```diff
@@ -6,127 +17,244 @@
...
+     */
+    while (1)
```
NataliaPavlovic: I think we should leave these as infinite for loops because the rest of the code uses them, and there is no difference in behaviour.

Src/ReadBarometer.c
```diff
@@ -6,127 +17,244 @@
...
     {
+        // Delay so that the loop operates at 50Hz
```
NataliaPavlovic: Frequency comment is unnecessary because it is already commented when READ_BAROMETER_PERIOD is declared.

Src/ReadBarometer.c
```diff
@@ -6,127 +17,244 @@
...
+
+        int32_t dT      = temperatureReading - (c5Tref << 8);
+        int32_t temp    = 2000 + ((dT * c6Tempsens) >> 23);                 // Divide this value by 100 to get degrees Celcius
+        int64_t off     = (c2Off << 17) + ((dT * c4Tco) >> 6);
+        int64_t sens    = (c1Sens << 16) + ((dT * c3Tcs) >> 7);
+        int32_t p       = (((pressureReading * sens) >> 21) - off) >> 15;   // Divide this value by 100 to get millibars
+
+        /* Calculate Second-Order Temperature and Pressure -------------------*/
+
+        // Equations provided by MS5607-02BA03 data sheet
```
NataliaPavlovic: Same comment as in line 182. Having both is redundant.

Src/ReadBarometer.c
```diff
@@ -6,127 +17,244 @@
...
-        data->pressure_ = (int32_t) pressure;
-        data->temperature_ = (int32_t) temperature;
-        osMutexRelease(data->mutex_);
+        // PRESSURE AND TEMPERATURE VALUES ARE STORED AT 100x VALUE TO MAINTAIN
+        // 2 DECIMAL POINTS OF PRECISION AS AN INTEGER!
+        // E.x. The value 1234 should be interpreted as 12.34
     }
 }
 
+/**
```
NataliaPavlovic, line 216: I like the idea of using javadocs style comments for functions. If this comment describes what the function does, we don't need to repeat it in the code. I would do it something like this.
```
/**
 * This function reads and returns a 16-bit coefficient from the barometer.
 * @param   PROM_READ_CMD   The command to send in order to read the desired
 *                          coefficient from PROM. See data sheet for commands.
 * @return                  The read coefficient.
 */
uint16_t readCalibrationCoefficient(const uint8_t PROM_READ_CMD)
{
    uint16_t    coefficient;
    uint8_t     dataInBuffer;

    HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_RESET);

    HAL_SPI_Transmit(&hspi2, &PROM_READ_CMD, CMD_SIZE, CMD_TIMEOUT);

    // Read the first byte (bits 15-8)
    HAL_SPI_TransmitReceive(&hspi2, &READ_BYTE_CMD, &dataInBuffer, CMD_SIZE, CMD_TIMEOUT);
    coefficient = dataInBuffer << 8;

    // Read the second byte (bits 7-0)
    HAL_SPI_TransmitReceive(&hspi2, &READ_BYTE_CMD, &dataInBuffer, CMD_SIZE, CMD_TIMEOUT);
    coefficient += dataInBuffer;

    HAL_GPIO_WritePin(BARO_CS_GPIO_Port, BARO_CS_Pin, GPIO_PIN_SET);

    return coefficient;
}
```

Src/ReadBarometer.c
Matt4890: We don't actually want to do that lol, it was just an example regarding expandability moving forward.

Src/ReadBarometer.c
Matt4890: Changed back to `for (;;)` for consistency

Src/ReadBarometer.c
Matt4890: Removed

Src/ReadBarometer.c
Matt4890: Removed

Src/ReadBarometer.c
Matt4890: Removed, and moved the remaining one to the end of the loop, as to be more general

Src/ReadBarometer.c
Matt4890: Code changed to match

Src/ReadBarometer.c
NataliaPavlovic: As long as we use ```temperatureReading = dataInBuffer << 16``` with an = sign, it should be the same.

Src/ReadBarometer.c
```diff
@@ -6,127 +19,223 @@
...
+
+            if (temp < TEMP_VERY_LOW)   // If the temperature is below -15°C
+            {
+                off2    = off2 + (15 * ((temp + 1500) * (temp + 1500)));
+                sens2   = sens2 + (8 * ((temp + 1500) * (temp + 1500)));
+            }
+
+            temp    = temp  - t2;
+            off     = off   - off2;
+            sens    = sens  - sens2;
```
NataliaPavlovic, line 194: I think p needs to be recalculated here with the new values.

Src/ReadBarometer.c
```diff
@@ -6,127 +19,226 @@
...
-        // pressure and temperature were not divided by 100 to keep the decimal places
-        double pressure = (pressureReading * sens / pow(2, 21) - off) / pow(2, 15); // need to divide P by 100 to get mbar
-        double temperature = (temp); // need to divide T by 100 to get degrees Celcius
+        /* Calculate First-Order Temperature and Pressure --------------------*/
+
+        int32_t dT      = temperatureReading - (c5Tref << 8);
+        int32_t temp    = 2000 + ((dT * c6Tempsens) >> 23); // Divide this value by 100 to get degrees Celcius
+        int64_t off     = (c2Off << 17) + ((dT * c4Tco) >> 6);
+        int64_t sens    = (c1Sens << 16) + ((dT * c3Tcs) >> 7);
+        // int32_t p    = (((pressureReading * sens) >> 21) - off) >> 15;
```
NataliaPavlovic: I think we can take these 2 comments out. If we wanted to get the first order pressure, it would make more sense to just comment out the calculation for the second order calculations and calculate the pressure below it.

Matt4890: Done

NataliaPavlovic (approved):
Changes are consistent with info on the datasheet.
Tested code. Logged pressure and temperature values were consistent with expected values.

Log file:
[VanderAvionics.csv.zip](https://github.com/StudentOrganisationForAerospaceResearch/AvionicsSoftware/files/2776206/VanderAvionics.csv.zip)

PR #6, opened 2019-03-16
logdata.c

missmarlos:
Check if there's a header before logging data onto the file.

### DIFF
--- a/Src/LogData.c
+++ b/Src/LogData.c
@@ -234,7 +234,6 @@ void logDataTask(void const* arg)
 
     for (;;)
     {
-        checkForHeader();
         if(containsHeader == 1){
             switch (getCurrentFlightPhase())
             {
@@ -261,28 +260,17 @@ void logDataTask(void const* arg)
                     break;
             }    
         }else if(containsHeader ==  0){
-            //Open file
+            //open the file
             f_open(&file, fileName, FA_CREATE_NEW | FA_READ | FA_WRITE);
-            //Write header
-            f_puts(buffer, &file);
-            //Make sure header is written to file
-            checkForHeader();
+            //if the file is empty, write a header
+            if(f_size(&file) == 0){
+                f_puts(buffer, &file);
+            //if there's something in the file, go to the switch statement
+            }else if(f_size(&file) == 1){
+                containsHeader == 1;
+            }
+            //close the file
+            f_close(&file);
         }
     }
-}
-
-//This function checks if there's a header
-void checkForHeader(){
-    //open file again then read if theres a header
-    f_open(&file, fileName, FA_CREATE_NEW | FA_READ | FA_WRITE);
-    //if the file is empty, keep trying to write header.
-    if(f_size(&file) == 0){
-        containsHeader = 0;
-    //if file is not empty, then the header is there.
-    //enter the switch statement to write rest of file.
-    }else{
-        containsHeader = 1;
-    }
-    //close the file
-    f_close(&file);
 }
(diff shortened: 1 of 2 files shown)


Src/hikingAlgorithm.java
```diff
@@ -0,0 +1,5 @@
+public class hikingAlgorithm{
```
NataliaPavlovic: This file must have been accidentally included. Delete this file from the project.

Src/LogData.c
```diff
@@ -16,6 +16,10 @@ static int FAST_LOG_DATA_PERIOD = 50;
 static FATFS fatfs;
 static FIL file;
 
+static int containsHeader = 0;
```
NataliaPavlovic, line 19: We can use a uin8_t for this variable instead of int because this data will always be 1 byte or smaller and will be unsigned. This variable can be a static variable placed above the for loop instead of being a global variable in this file. Static variables in functions retain their value over all calls to that function.

Src/LogData.c
```diff
@@ -230,29 +234,55 @@ void logDataTask(void const* arg)
...
-
-            case ABORT:
-                lowFrequencyLogToSdRoutine(data, buffer, ABORT);
-                break;
-
-            default:
-                lowFrequencyLogToSdRoutine(data, buffer, MAIN_DESCENT); // won't work
-                break;
+        checkForHeader();
+        if(containsHeader == 1){
```
NataliaPavlovic, line 237: Move ```{``` to new line

Src/LogData.c
```diff
@@ -230,29 +234,55 @@ void logDataTask(void const* arg)
...
+
+                case ABORT:
+                    lowFrequencyLogToSdRoutine(data, buffer, ABORT);
+                    break;
+
+                default:
+                    lowFrequencyLogToSdRoutine(data, buffer, MAIN_DESCENT); // won't work
+                    break;
+            }    
+        }else if(containsHeader ==  0){
```
NataliaPavlovic, line 262: Move else if to new line and ```{``` to another new line

Src/LogData.c
```diff
@@ -230,29 +234,55 @@ void logDataTask(void const* arg)
...
         }
     }
 }
+
+//This function checks if there's a header
+void checkForHeader(){
+    //open file again then read if theres a header
+    f_open(&file, fileName, FA_CREATE_NEW | FA_READ | FA_WRITE);
+    //if the file is empty, keep trying to write header.
+    if(f_size(&file) == 0){
```
NataliaPavlovic: Since containsHeader starts at 0, we can replace this ```if else``` statement with a single if statement.
```
if(f_size(&file))
{
    containsHeader = 1;
}
```
This will only change containsHeader from 0 to 1 if the header is written.

Src/LogData.c
```diff
@@ -230,29 +234,55 @@ void logDataTask(void const* arg)
...
-                break;
-
-            case ABORT:
-                lowFrequencyLogToSdRoutine(data, buffer, ABORT);
-                break;
-
-            default:
-                lowFrequencyLogToSdRoutine(data, buffer, MAIN_DESCENT); // won't work
-                break;
+        checkForHeader();
```
NataliaPavlovic: I think it is unnecessary to checkForHeader every time we go through this loop. I think this one can be removed, and then we can just check for it in the ```else if(containsHeader ==  0)```

NataliaPavlovic (changes requested):
Once the comments are addressed, can you push your code to this branch again. I will get an email notification and then I can review it some more.

PR #9, opened 2019-03-23
Changing the radio baud rate to 9600

sudamf1:


NataliaPavlovic (approved):
Baud Rate changed to what the radio was configured to. Project builds.

PR #15, opened 2019-04-02
Natalia/v3 pin configurations

NataliaPavlovic:


NataliaPavlovic:
Verified that none of the logging or collected data is affected by this change.

PR #20, opened 2019-05-26
Natalia/change sol5 pin

NataliaPavlovic:


NataliaPavlovic:
Tested during soft mock launch. Verified.

PR #24, opened 2019-07-19
Lowered BURN duration to 8.5s

NataliaPavlovic:


NataliaPavlovic:
Correction: This sets the duration to 8.5s

PR #25, opened 2019-11-16
Natalia/gps 2.0

NataliaPavlovic:
I have tested this code with an SD card, and the data is correct. I have not tested it with Ground Systems and radio communication yet.

I will also add more comments into the GPS code soon.


Src/ReadGps.c
```diff
@@ -6,15 +6,131 @@
...
+                    {
+                        data->time_ = (uint32_t) (atof(gps_item) * 100); // HHMMSS.SS format. Time is multiplied by 100.
+                        break;
+                    }
+
+                    case 2:
+                    {
+                        double latitude = (atof(gps_item)); // DDMM.MMMMMM
+                        data->latitude_.degrees_ = (int32_t) latitude / 100; // First 2 numbers are the latitude degrees
+                        data->latitude_.minutes_ = (uint32_t) ((latitude - data->latitude_.degrees_ * 100) * 100000); // Latitude minutes is multplied by 100000
```
NataliaPavlovic: Change to int32_t because it should be signed.

Src/ReadGps.c
```diff
@@ -6,15 +6,131 @@
...
+                        }
+
+                        break;
+                    }
+
+                    case 4:
+                    {
+                        double longitude = (atof(gps_item)); // DDMM.MMMMMM
+                        data->longitude_.degrees_ = (int32_t) longitude / 100; // First 2 numbers are the longitude degrees
+                        data->longitude_.minutes_ = (uint32_t) ((longitude - data->longitude_.degrees_ * 100) * 100000); // Longitude minutes is multplied by 100000
```
NataliaPavlovic: Change to int32_t because it should be signed.

NataliaPavlovic:
Verified that the correct data is received from the GPS and that this data can be transmitted over the radio.

Src/TransmitData.c
```diff
@@ -30,7 +34,7 @@ static const int8_t LOWER_VALVE_STATUS_HEADER_BYTE = 0x39;
 #define FLAGS_AND_CRC_SIZE (6) //1 byte for header flag, 1 byte for ender flag, 4 bytes for crc
 #define IMU_SERIAL_MSG_SIZE (37)
 #define BAROMETER_SERIAL_MSG_SIZE (9)
-#define GPS_SERIAL_MSG_SIZE (17)
+#define GPS_SERIAL_MSG_SIZE (25)
```
Abhay2412, line 37: Just wondering why this was changed from 17 to 25?

NataliaPavlovic: We put 17 there are a place holder because we didn't actually know how large the GPS data would actually be before we got the GPS working.
With the new communication protocol, we need 25 Bytes for the message (header byte + actual message).
- Byte 1: Header Byte (1 B)
- Bytes 2-5: gps_time (4 B)
- Bytes 6-9: latitude_degrees (4 B)
- Bytes 10-13: latitude_minutes (4 B)
- Bytes 14-17: longitude_degrees (4 B)
- Bytes 18-21: longitude_minutes (4 B)
- Bytes 22-25: altitude (4 B long)

Matt4890 (approved):
Looks Great! I'll need to update myself on how long/lat minutes work, but that'll be simple enough.

PR #32, opened 2021-04-17
Agam/andromeda v3.2

AgamAulakh:
updated based on kevin's comments :D ready for review #2

### DIFF
--- a/Src/LogData.c
+++ b/Src/LogData.c
@@ -1,3 +1,12 @@
+/**
+ ******************************************************************************
+ * File Name          : LogData.c
+ * Description        : Code handles logging of AllData as per FlightPhase
+ ******************************************************************************
+*/
+
+
+/* Includes ------------------------------------------------------------------*/
 #include "stm32f4xx.h"
 #include "stm32f4xx_hal_conf.h"
 #include "cmsis_os.h"
@@ -10,14 +19,8 @@
 #include "Data.h"
 #include "FlightPhase.h"
 
-static int SLOW_LOG_DATA_PERIOD = 700;
-static int FAST_LOG_DATA_PERIOD = 200;
-static uint8_t softwareVersion = 104;
-static uint8_t deviceAddress = 0x50;
-static uint8_t currMemAddress = 0x07;
-static uint32_t timeout = 10;//what sbould this be?
-//NOTE: the functions pass around a timeout but i made a static declaration here
 
+/* Structs -------------------------------------------------------------------*/
 typedef struct{
     int32_t accelX;
     int32_t accelY;
@@ -37,21 +40,30 @@ typedef struct{
     int32_t latitude_minutes;
     int32_t longitude_degrees;
     int32_t longitude_minutes;
+    int32_t antennaAltitude;
+    int32_t geoidAltitude;
     int32_t altitude;
-} logEntry;
+    int32_t tick;
+} logEntry; // logEntry holds data from AllData that is to be logged
 
 
-//SUPER COOL THING BELOW!
-static logEntry theLogEntry;
+/* Constants -----------------------------------------------------------------*/
+static int SLOW_LOG_DATA_PERIOD = 700;
+static int FAST_LOG_DATA_PERIOD = 200;
+static uint8_t softwareVersion = 104;
+static uint8_t deviceAddress = 0x50;
+static uint8_t currMemAddress = 0x07; //TODO: update this as needed
+static uint32_t timeout = 10; // Time required to wait before ending read/write 
+static logEntry theLogEntry; // This is the global log entry that is updated
+static uint8_t logEntrySize = 80;
 
-
+/* Functions -----------------------------------------------------------------*/
 /**
  * @brief Writes data to the EEPROM over I2C.
  * @param buffer, pointer to the data buffer.
  * @param bufferSize, size of data buffer.
- * @param timeout, time to wait for write operation before timing out in ms.
  */
-void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, uint16_t timeout)
+void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress)
 {
 	HAL_I2C_Mem_Write(&hi2c1, deviceAddress<<1, memAddress, uint16_t(sizeof(memAddress)), buffer, bufferSize, timeout);
 }
@@ -60,18 +72,16 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
  * @brief Reads data from the EEPROM over I2C.
  * @param receiveBuffer, pointer to the buffer to store received data.
  * @param bufferSize, size of data to be read.
- * @param timeout, time to wait for read operation before timing out in ms.
  */
-void readFromEEPROM(uint8_t* receiveBuffer, uint16_t bufferSize, uint16_t memAddress, uint16_t timeout)
+void readFromEEPROM(uint8_t* receiveBuffer, uint16_t bufferSize, uint16_t memAddress)
 {
 	HAL_I2C_Mem_Read(&hi2c1, deviceAddress<<1, memAddress, sizeof(memAddress), receiveBuffer, bufferSize, timeout);
 }
 
 /**
  * @brief Blocks the current thread until the I2C state is ready for reading/writing.
- * @param timeout Timeout duration
  */
-void checkEEPROMBlocking(uint16_t timeout)
+void checkEEPROMBlocking()
 {
 	while(HAL_I2C_GetState(&hi2c1) != HAL_I2C_STATE_READY){};
 	while (HAL_I2C_IsDeviceReady(&hi2c1, deviceAddress<<1, 3, timeout) != HAL_OK){};
@@ -79,12 +89,11 @@ void checkEEPROMBlocking(uint16_t timeout)
 
 /**
  * @brief puts a logEntry into a char array and sends to EEPROM for writing
- * @param memAddress, needed by the writeToEEPROM function
- * @param timeout, time to wait for write operation
+ * @param memAddress, the address at which the EEPROM is instructed to start writing
  */
-void writeLogEntryToEEPROM(uint16_t memAddress, uint16_t timeout)
+void writeLogEntryToEEPROM(uint16_t memAddress)
 {
-    char dataToWrite[200]; 
+    char dataToWrite[logEntrySize]; 
 
     dataToWrite[0] = theLogEntry.accelX;
     dataToWrite[4] = theLogEntry.accelY;
@@ -95,39 +104,34 @@ void writeLogEntryToEEPROM(uint16_t memAddress, uint16_t timeout)
     dataToWrite[24] = theLogEntry.magnetoX;
     dataToWrite[28] = theLogEntry.magnetoY;
     dataToWrite[32] = theLogEntry.magnetoZ;
-
     dataToWrite[36] = theLogEntry.barometerPressure;
     dataToWrite[40] = theLogEntry.barometerTemperature;
-
     dataToWrite[44] = theLogEntry.combustionChamberPressure;
     dataToWrite[48] = theLogEntry.oxidizerTankPressure;
-
     dataToWrite[52] = theLogEntry.gps_time;
     dataToWrite[56] = theLogEntry.latitude_degrees;
     dataToWrite[60] = theLogEntry.latitude_minutes;
     dataToWrite[64] = theLogEntry.longitude_degrees;
     dataToWrite[68] = theLogEntry.longitude_minutes;
     dataToWrite[72] = theLogEntry.altitude;
+    dataToWrite[76] = theLogEntry.tick;
 
-    checkEEPROMBlocking(timeout);
-    writeToEEPROM(dataToWrite, 76, memAddress, timeout);
-    // need: memAddress and timeout??????
-    // increment address
+    checkEEPROMBlocking();
+    writeToEEPROM(dataToWrite, sizeof(dataToWrite), memAddress);
+    //TODO: make sure memAddress is correct, increment address
 }
 
 /**
  * @brief reads logEntry-sized char array from EEPROM and updates theLogEntry
- * @param addr, I dont know where this is used
- * @param memAddress, needed by the readFromEEPROM function
- * @param timeout, time to wait for read operation
+ * @param memAddress, the address at which the EEPROM is instructed to start reading
  */
-void readLogEntryToEEPROM(uint16_t addr, uint16_t memAddress, uint16_t timeout)
+void readLogEntryToEEPROM(uint16_t memAddress)
 {
-    char dataRead[200];
+    char dataRead[logEntrySize];
   
-    checkEEPROMBlocking(timeout);
-    readFromEEPROM(dataRead, 80, memAddress, timeout);
-    // need: memAdress and timeout
+    checkEEPROMBlocking();
+    readFromEEPROM(dataRead, sizeof(dataRead), memAddress);
+    //TODO: memAdress may not be corrrect bc it's never incremented
 
     theLogEntry.accelX = dataRead[0];
     theLogEntry.accelY = dataRead[4];
@@ -138,25 +142,25 @@ void readLogEntryToEEPROM(uint16_t addr, uint16_t memAddress, uint16_t timeout)
     theLogEntry.magnetoX = dataRead[24];
     theLogEntry.magnetoY = dataRead[28];
     theLogEntry.magnetoZ = dataRead[32];
-
     theLogEntry.barometerPressure = dataRead[36];
     theLogEntry.barometerTemperature = dataRead[40];
-
     theLogEntry.combustionChamberPressure = dataRead[44];
     theLogEntry.oxidizerTankPressure = dataRead[48];
-
     theLogEntry.gps_time = dataRead[52];
     theLogEntry.latitude_degrees = dataRead[56];
     theLogEntry.latitude_minutes = dataRead[60];
     theLogEntry.longitude_degrees = dataRead[64];
     theLogEntry.longitude_minutes = dataRead[68];
     theLogEntry.altitude = dataRead[72];
+    theLogEntry.tick = dataRead[76];
 }
 
 /**
- * @brief
+ * @brief this method initializes the log entry struct
+ * and is only called during the start of the task
  */
-void initializeLogEntry(){
+void initializeLogEntry()
+{
     theLogEntry.accelX = -1;
     theLogEntry.accelY = -1;
     theLogEntry.accelZ = -1;
@@ -169,22 +173,24 @@ void initializeLogEntry(){
     theLogEntry.barometerPressure = -1;
     theLogEntry.barometerTemperature = -1;
     theLogEntry.combustionChamberPressure = -1;
-    theLogEntry.oxidizerTankPressure = -1; //this too
-    // GPS
+    theLogEntry.oxidizerTankPressure = -1;
     theLogEntry.gps_time = 0xFFFF;
     theLogEntry.latitude_degrees = -1;
     theLogEntry.latitude_minutes = 0xFFFF;
     theLogEntry.longitude_degrees = -1;
     theLogEntry.longitude_minutes = 0xFFFF;
+    theLogEntry.antennaAltitude = -1;
+    theLogEntry.geoidAltitude = -1;
     theLogEntry.altitude = -1;
+    theLogEntry.tick = HAL_GetTick();
 }
 /**
- * @brief Simply updates theLogEntry field with CURRENT values in AllData struct, formerly buildLogEntry
+ * @brief Simply updates theLogEntry field with CURRENT values in AllData struct
  * @param data, pointer to AllData struct
  */
-void updateLogEntry(AllData* data)//updateLogEntry, pointer to the log entry
+void updateLogEntry(AllData* data)
 {
-    if (osMutexWait(data->accelGyroMagnetismData_->mutex_, 0) == osOK) //this is how we do it!
+    if (osMutexWait(data->accelGyroMagnetismData_->mutex_, 0) == osOK)
     {
         theLogEntry.accelX = data->accelGyroMagnetismData_->accelX_;
         theLogEntry.accelY = data->accelGyroMagnetismData_->accelY_;
@@ -220,28 +226,28 @@ void updateLogEntry(AllData* data)//updateLogEntry, pointer to the log entry
     if (osMutexWait(data->gpsData_->mutex_, 0) == osOK)
     {
         theLogEntry.gps_time = data->gpsData_->time_;
-
         theLogEntry.latitude_degrees = data->gpsData_->latitude_.degrees_;
         theLogEntry.latitude_minutes = data->gpsData_->latitude_.minutes_;
-
         theLogEntry.longitude_degrees = data->gpsData_->longitude_.degrees_;
         theLogEntry.longitude_minutes = data->gpsData_->longitude_.minutes_;
-
+        theLogEntry.antennaAltitude = data->gpsData_->totalAltitude_.altitude_;
+        theLogEntry.geoidAltitude = data->gpsData_->totalAltitude_.altitude_;
         theLogEntry.altitude = data->gpsData_->totalAltitude_.altitude_;
-
         osMutexRelease(data->gpsData_->mutex_);
     }
+    //TODO: add FlightPhase to logEntry and update it
+    theLogEntry.tick = HAL_GetTick();
 }
 
 /**
- * Used for logging entries to the EEPROM. First updates the struct, then writes it
+ * @brief Used for logging entries to the EEPROM. First updates the struct, then writes it
  * the wait times are all managed in the main for loop so highFrequencyLog = lowFrequencyLog
- * @param data, sent to updateLogEntry
+ * @param data, pointer to AllData Struct sent to updateLogEntry to update the logEntry struct
  */
 void logEntryOnceRoutine(AllData* data)
 {
     updateLogEntry(data);
-    writeLogEntryToEEPROM(currMemAddress, timeout);
+    writeLogEntryToEEPROM(currMemAddress);
 }
 
 void logDataTask(void const* arg)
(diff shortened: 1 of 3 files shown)


Src/LogData.c
```diff
@@ -10,24 +19,52 @@
...
+    int32_t currentFlightPhase;
+    int32_t tick;
+} logEntry; // logEntry holds data from AllData that is to be logged
+
+
+/* Constants -----------------------------------------------------------------*/
 static int SLOW_LOG_DATA_PERIOD = 700;
 static int FAST_LOG_DATA_PERIOD = 200;
 static uint8_t softwareVersion = 104;
 static uint8_t deviceAddress = 0x50;
```
zelongzhang: Need comments for globals/statics. PERIODS in milliseconds, should be in name.

Src/LogData.c
```diff
@@ -10,24 +19,52 @@
...
-static uint8_t currMemAddress = 0x07;
-
-static FATFS fatfs;
-static FIL file;
-
-char fileName[32];
+static uint8_t currMemAddress = 0x07; //TODO: update this as needed
+static uint32_t timeout = 10; //Time required to wait before ending read/write (unit =?)
+static logEntry theLogEntry; //This is the global log entry that is updated
+static uint8_t logEntrySize = 84;
```
zelongzhang: Need comments for globals/statics

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
+    theLogEntry.barometerTemperature = dataRead[40];
+    theLogEntry.combustionChamberPressure = dataRead[44];
+    theLogEntry.oxidizerTankPressure = dataRead[48];
+    theLogEntry.gps_time = dataRead[52];
+    theLogEntry.latitude_degrees = dataRead[56];
+    theLogEntry.latitude_minutes = dataRead[60];
+    theLogEntry.longitude_degrees = dataRead[64];
+    theLogEntry.longitude_minutes = dataRead[68];
+    theLogEntry.altitude = dataRead[72];
+    theLogEntry.currentFlightPhase = dataRead][76];
```
zelongzhang: Extra "]" char.

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
+        theLogEntry.magnetoZ = data->accelGyroMagnetismData_->magnetoZ_;
         osMutexRelease(data->accelGyroMagnetismData_->mutex_);
     }
 
     if (osMutexWait(data->barometerData_->mutex_, 0) == osOK)
     {
-        pressure = data->barometerData_->pressure_;
-        temperature = data->barometerData_->temperature_;
+        theLogEntry.pressure = data->barometerData_->pressure_;
+        theLogEntry.temperature = data->barometerData_->temperature_;
```
zelongzhang: theLogEntry.pressure and theLogEntry.temperature are not declared.

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
-    );
-
+    initializeLogEntry();
+    uint32_t prevWakeTime;
+    clock_t beforeLogTime, afterLogTime, totalTime;
     for (;;)
     {
+        beforeLogTime = clock();
+        logEntryOnceRoutine();
+        afterLogTime = clock();
```
zelongzhang: Let's use osKernelSysTick() here, it's suppose to return a time in milliseconds as uint_32. This also means that totalTime can be calculated simply as
afterLogTime - beforeLogTime
since they're both in milliseconds already and so is our LOG PERIOD consts.

With this change, we also won't need to include time.h

Src/LogData.c
```diff
@@ -10,24 +19,52 @@
...
+    int32_t longitude_minutes;
+    int32_t antennaAltitude;
+    int32_t geoidAltitude;
+    int32_t altitude;
+    int32_t currentFlightPhase;
+    int32_t tick;
+} logEntry; // logEntry holds data from AllData that is to be logged
+
+
+/* Constants -----------------------------------------------------------------*/
```
Matt4890, line 59: Constants needs the `const` keyword in order to be constant

Src/LogData.c
```diff
@@ -10,24 +19,52 @@
...
 static int SLOW_LOG_DATA_PERIOD = 700;
 static int FAST_LOG_DATA_PERIOD = 200;
```
Matt4890: Try to use more specific types (`uint8_t`, `int32_t`, etc)

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
+    theLogEntry.gps_time = 0xFFFF;
+    theLogEntry.latitude_degrees = -1;
+    theLogEntry.latitude_minutes = 0xFFFF;
+    theLogEntry.longitude_degrees = -1;
+    theLogEntry.longitude_minutes = 0xFFFF;
+    theLogEntry.antennaAltitude = -1;
+    theLogEntry.geoidAltitude = -1;
+    theLogEntry.altitude = -1;
+    theLogEntry.currentFlightPhase = getCurrentFlightPhase();
+    theLogEntry.tick = HAL_GetTick();
```
AgamAulakh: use osKernelSysTick();

zelongzhang (changes requested):

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
+        afterLogTime = clock();
+        totalTime = (double)(afterLogTime - beforeLogTime) / CLOCKS_PER_SEC / 1000; //this also takes time!
+        prevWakeTime = osKernelSysTick();//assume it is atomic: runs in one cycle
         switch (getCurrentFlightPhase())
         {
             case BURN:
             case COAST:
             case DROGUE_DESCENT:
-                highFrequencyLogToSdRoutine(data, buffer);
+                osDelayUntil(&prevWakeTime, (uint32_t)(FAST_LOG_DATA_PERIOD - totalTime));
```
Matt4890: Needs check for `totalTime > log period`. Otherwise negative time will be cast to near max uint32 size (500+ million years). Suggested solution: `(uint32_t)(max(LOG_PERIOD - totalTime, 0))`.

Also, to be more accurate, we should call the second time-get-function in the delay: `(uint32_t)(max(LOG_PERIOD - (osKernelSysTick() - beforeLogTime), 0))`. Don't need to worry about the clock rolling over, since that needs the clock to be running for 500+ million years.

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
-        "elapsedTime(ms),"
-        "softwareVersion\n"
-    );
-
+    initializeLogEntry();
+    uint32_t prevWakeTime;
+    clock_t beforeLogTime, afterLogTime, totalTime;
     for (;;)
     {
+        beforeLogTime = clock();
```
Matt4890: Need to figure out the difference between `clock()`, `osKernelSysTick()`, and `HAL_GetTick()`. I would suggest `osKernelSysTick()` for now.

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
-        {
-            f_puts(buffer, &file);
-            f_close(&file); // close to save the file
-        }
-    }
-
-    HAL_GPIO_WritePin(LED1_GPIO_Port, LED1_Pin, 0);
-    f_mount(NULL, "SD:", 1);
+    updateLogEntry(data);
+    writeLogEntryToEEPROM(currMemAddress);
```
Matt4890: Need to increment offset:
```
writeLogEntryToEEPROM(EEPROM_START_ADDRESS + logAddressOffset);
logAddressOffset += LOG_SIZE_BYTES;
```

Need to include space-saving logic here. Probably needs to be in a different change.

Src/LogData.c
```diff
@@ -10,24 +19,52 @@
...
 static int FAST_LOG_DATA_PERIOD = 200;
 static uint8_t softwareVersion = 104;
 static uint8_t deviceAddress = 0x50;
-static uint8_t currMemAddress = 0x07;
-
-static FATFS fatfs;
-static FIL file;
-
-char fileName[32];
+static uint8_t currMemAddress = 0x07; //TODO: update this as needed
```
Matt4890: Should change to a `EEPROM_START_ADDRESS` constant and a `logAddressOffset` variable.

Src/LogData.c
Matt4890: Don't know if `max(...)` is currently implemented. Otherwise we can use the following:
```
 #define max(a,b) \
   ({ __typeof__ (a) _a = (a); \
       __typeof__ (b) _b = (b); \
     _a > _b ? _a : _b; })
 ```
https://stackoverflow.com/questions/3437404/min-and-max-in-c

Comment (Matt4890) [reply Src/LogData.c]: Or find a better way 🤷

File: Src/LogData.c [hunk 13]
```diff
@@ -10,24 +19,52 @@
...
+static uint32_t timeout = 10; //Time required to wait before ending read/write (unit =?)
```
Matt4890: Unit = milliseconds most likely

Src/LogData.c
Matt4890: Can we not use `sizeof(logEntry)` instead of having a constant? The `sizeof(...)` function gets evaluated at compile time, so no speed penalty there. Also, unless some of the types in the `logData` struct need to be changed, then this value is wrong (i.e. a `logData` struct is currently 23*uint32_t, which is 96 bytes).

Src/LogData.c
```diff
@@ -10,24 +19,52 @@
...
+    int32_t latitude_degrees;
+    int32_t latitude_minutes;
+    int32_t longitude_degrees;
+    int32_t longitude_minutes;
+    int32_t antennaAltitude;
+    int32_t geoidAltitude;
+    int32_t altitude;
+    int32_t currentFlightPhase;
+    int32_t tick;
+} logEntry; // logEntry holds data from AllData that is to be logged
```
Matt4890: Type (and struct) names should begin with an uppercase letter: `LogEntry`

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
 
 /**
  * @brief Blocks the current thread until the I2C state is ready for reading/writing.
- * @param timeout Timeout duration
  */
-void checkEEPROMBlocking(uint16_t timeout)
+void checkEEPROMBlocking()
 {
 	while(HAL_I2C_GetState(&hi2c1) != HAL_I2C_STATE_READY){};
 	while (HAL_I2C_IsDeviceReady(&hi2c1, deviceAddress<<1, 3, timeout) != HAL_OK){};
```
Matt4890: We should maybe add a small (~100us?) delay in these spinlocks. It will noticeably reduce power consumption iirc. Thoughts? Also interested if the compiler would optimize these out.

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
 {
 	HAL_I2C_Mem_Read(&hi2c1, deviceAddress<<1, memAddress, sizeof(memAddress), receiveBuffer, bufferSize, timeout);
 }
 
 /**
  * @brief Blocks the current thread until the I2C state is ready for reading/writing.
- * @param timeout Timeout duration
  */
-void checkEEPROMBlocking(uint16_t timeout)
+void checkEEPROMBlocking()
```
Matt4890, line 97: Would `wait/blockUntilEEPROMReady(...)` be a better name?

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
+    dataToWrite[36] = theLogEntry.barometerPressure;
+    dataToWrite[40] = theLogEntry.barometerTemperature;
+    dataToWrite[44] = theLogEntry.combustionChamberPressure;
+    dataToWrite[48] = theLogEntry.oxidizerTankPressure;
+    dataToWrite[52] = theLogEntry.gps_time;
+    dataToWrite[56] = theLogEntry.latitude_degrees;
+    dataToWrite[60] = theLogEntry.latitude_minutes;
+    dataToWrite[64] = theLogEntry.longitude_degrees;
+    dataToWrite[68] = theLogEntry.longitude_minutes;
+    dataToWrite[72] = theLogEntry.altitude;
```
Matt4890: Missing `antennaAltitude` and `geoidAltitude`. Is there a reason we wouldn't want to log these?

Matt4890: See next comment

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
+    dataToWrite[76] = theLogEntry.currentFlightPhase;
+    dataToWrite[80] = theLogEntry.tick;
+    checkEEPROMBlocking();
+    writeToEEPROM(dataToWrite, sizeof(dataToWrite), memAddress);
```
Matt4890: Structs are contiguous blocks of memory, so you could instead cast `theLogEntry` to a `char*` and pass it directly to `writeToEEPROM(...)` without needing `dataToWrite`:
```c
writeToEEPROM((char*)theLogEntry, sizeof(LogEntry), memAddress);
```

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
+    //TODO: make sure memAddress is correct, increment address
```
Matt4890: Should probably do this in this commit too

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
+    theLogEntry.combustionChamberPressure = dataRead[44];
+    theLogEntry.oxidizerTankPressure = dataRead[48];
+    theLogEntry.gps_time = dataRead[52];
+    theLogEntry.latitude_degrees = dataRead[56];
+    theLogEntry.latitude_minutes = dataRead[60];
+    theLogEntry.longitude_degrees = dataRead[64];
+    theLogEntry.longitude_minutes = dataRead[68];
+    theLogEntry.altitude = dataRead[72];
+    theLogEntry.currentFlightPhase = dataRead][76];
+    theLogEntry.tick = dataRead[80];
```
Matt4890: Why are we storing the data in `theLogEntry`?
We should probably be using an "output" argument here, which is simply a pointer to data that the caller allocates, which this function would "output" to. Example:
```c
/**
 * @brief Initializes a LogEntry variable to all 1's.
 * @param[in] output    Pointer to the LogEntry struct to initialize.
 */
void initLogEntry(LogEntry* output) {
    // Should replace this loop with a memcpy(...)
    for (int i = 0; i < sizeof(LogEntry); i++) {
        // Set every byte to all 1's
        ((char*)output)[i] = 0xFF; // = 255 = 0b11111111
    }
}

// Use case
int main() {
    ...
    LogEntry log;
    // Log currently uninitialized (maybe all 0's, maybe random garbage data)

    initLogEntry(&log);
    // Log entry now all 1's (each byte is 0xFF = 0b11111111)
    ...
}
```
This way the calling function has control over the allocation of the data. If we use these kinds of arguments, then we can probably remove the global variable in its entirety.

If you're wondering why we don't just declare the variable locally within the function and return it, then think about the following: 
- What would happen if the argument wasn't a struct, but rather an array (which is really just a pointer).
- What type of memory the array data is stored in if we declare it locally.
- What actually gets returned.

This was a recruiting question I got asked at the career fair in 2019/2020 by Synopsis. Definitely give it a good few minutes of thought before googling it, or ask me and I'll explain it :) (it's actually pretty interesting)

Src/LogData.c
Matt4890: Could probably replace this with a single `memcpy(...)` (See http://www.cplusplus.com/reference/cstring/memcpy/ if you're unfamiliar with it) setting the entire entry to -1 (0xFF in every byte), followed by the two last lines that set actual values.

i.e.
```c
    memcpy(/* I'll let you figure out the args :) */);
    theLogEntry.currentFlightPhase = getCurrentFlightPhase();
    theLogEntry.tick = osKernelSysTick();
```

Src/LogData.c
```diff
@@ -36,316 +73,209 @@ void writeToEEPROM(uint8_t* buffer, uint16_t bufferSize, uint16_t memAddress, ui
...
+        theLogEntry.latitude_minutes = data->gpsData_->latitude_.minutes_;
+        theLogEntry.longitude_degrees = data->gpsData_->longitude_.degrees_;
+        theLogEntry.longitude_minutes = data->gpsData_->longitude_.minutes_;
+        theLogEntry.antennaAltitude = data->gpsData_->antennaAltitude_.altitude_;
+        theLogEntry.geoidAltitude = data->gpsData_->geoidAltitude_.altitude_;
+        theLogEntry.altitude = data->gpsData_->totalAltitude_.altitude_;
+        osMutexRelease(data->gpsData_->mutex_);
     }
+    theLogEntry.currentFlightPhase = getCurrentFlightPhase();
+    theLogEntry.tick = HAL_GetTick();
```
Matt4890: Use `osKernelSysTick()`.

Matt4890 (changes requested):
Sorry for the wall of comments lol. Not all of the comments are code you specifically wrote, but are things we should change anyways to make the file cleaner. Feel free to reach out to either Kevin or I if you have any questions!

Src/LogData.c
```diff
@@ -45,18 +52,19 @@ typedef struct{
...
-static uint8_t deviceAddress = 0x50;
-static uint8_t currMemAddress = 0x07; //TODO: update this as needed
-static uint32_t timeout = 10; //Time required to wait before ending read/write (unit =?)
-static logEntry theLogEntry; //This is the global log entry that is updated
-static uint8_t logEntrySize = 84;
+static const int32_t SLOW_LOG_DATA_PERIOD_ms = 700; // logging period for slow routine
+static const int32_t FAST_LOG_DATA_PERIOD_ms = 200; // logging period for fast routine
+static const uint8_t softwareVersion = 104; // this is not used at all
+static const uint8_t deviceAddress = 0x50; // used for reading/writing to EEPROM
+static uint8_t EEPROM_START_ADDRESS = 0x07; // start address for reading/writing
```
Matt4890: `EEPROM_START_ADDRESS` should be `const`

Src/LogData.c
```diff
@@ -45,18 +52,19 @@ typedef struct{
...
+static uint8_t logAddressOffset = 0; // offset updated after writing in logEntryOnceRoutine
+static const uint32_t timeout_ms = 10; // Time required to wait before ending read/write
+static LogEntry theLogEntry; // This is the global log entry that is updated
```
Matt4890: We should move the non-constants under a new "Variables" section

Src/LogData.c
```diff
@@ -45,18 +52,19 @@ typedef struct{
...
-static uint32_t timeout = 10; //Time required to wait before ending read/write (unit =?)
-static logEntry theLogEntry; //This is the global log entry that is updated
-static uint8_t logEntrySize = 84;
+static const int32_t SLOW_LOG_DATA_PERIOD_ms = 700; // logging period for slow routine
+static const int32_t FAST_LOG_DATA_PERIOD_ms = 200; // logging period for fast routine
+static const uint8_t softwareVersion = 104; // this is not used at all
+static const uint8_t deviceAddress = 0x50; // used for reading/writing to EEPROM
+static uint8_t EEPROM_START_ADDRESS = 0x07; // start address for reading/writing
+static uint8_t logAddressOffset = 0; // offset updated after writing in logEntryOnceRoutine
+static const uint32_t timeout_ms = 10; // Time required to wait before ending read/write
```
Matt4890: We should have constants use `UPPERCASE_AND_UNDERSCORES` naming for consistency and clarity

Src/LogData.c
Matt4890: @AgamAulakh @zelongzhang Thoughts?

Src/LogData.c
```diff
@@ -45,18 +52,19 @@ typedef struct{
...
+static LogEntry theLogEntry; // This is the global log entry that is updated
+static const uint8_t LogEntrySize = sizeof(LogEntry); //size of logEntry = 96 bytes
```
Matt4890: I would actually just use `sizeof(LogEntry)` instead of having the `LogEntrySize` constant at all. For example on line 141: `char dataRead[sizeof(LogEntry)];`

Src/LogData.c
Matt4890: Bump

Src/LogData.c
```diff
@@ -184,8 +195,14 @@ void initializeLogEntry()
     theLogEntry.antennaAltitude = -1;
     theLogEntry.geoidAltitude = -1;
     theLogEntry.altitude = -1;
+    //[1] Using memcpy: does it only update one element of the struct?
+    //[2] when using memcpy to update entire struct, would we need a temp stuct?
+
+    //because struct memory is contiguous, is this correct:
+    //char initValue = 0xFF;
+    //memcpy (&LogEntry, &initValue, sizeof(LogEntry));
```
Matt4890: Whoops, I meant `memset(...)`, not `memcpy(...)`. The difference is that `memset(...)` *sets* a certain number of bytes at a pointer to a distinct value, and `memcpy(...)` *copies* a certain number of bytes from one pointer to another.

Right now, what would happen is that it would copy `sizeof(LogEntry)` number of bytes from the memory address of `initValue` to the memory address of `theLogEntry` (Careful of the typo there, you want to copy to the variable, not the type). The issue is that `initValue` is a `uint8_t`, i.e. only 1 byte, and a `LogEntry` variable is 90+ bytes. `memcpy(...)` will not "repeat" the value pointed to in the second argument, but will instead keep reading memory after where `initValue` is defined, which could include other variables, uninitialized garbage data, or areas of memory that aren't allowed to be accessed, which would throw a `Segmentation Fault`, causing the entire system to crash.

Here's the *right* reference page lol: http://www.cplusplus.com/reference/cstring/memset/

Src/LogData.c
```diff
@@ -250,32 +267,32 @@ void updateLogEntry(AllData* data)
 void logEntryOnceRoutine(AllData* data)
 {
     updateLogEntry(data);
-    writeLogEntryToEEPROM(currMemAddress);
+    writeLogEntryToEEPROM(EEPROM_START_ADDRESS + logAddressOffset);
+    logAddressOffset += LogEntrySize;
```
Matt4890: Just use `sizeof(LogEntry)` here.

Src/LogData.c
```diff
@@ -250,32 +267,32 @@ void updateLogEntry(AllData* data)
...
 }
 
 void logDataTask(void const* arg)
 {
     AllData* data = (AllData*) arg;
+    //TODO: make log entry here and pass it around (LogEntry log;) instead of a global var?
```
Matt4890: Yup! Global variables aren't always a good idea, so something like this would be better:
```c
LogEntry log;
initializeLogEntry(&log);
```

Matt4890 (changes requested):
Some comments on the left side might have been missed last time. They're kinda hidden from all the focus lol

Src/LogData.c
```diff
@@ -10,342 +19,280 @@
...
+    int32_t geoidAltitude;
+    int32_t altitude;
+    int32_t currentFlightPhase;
+    int32_t tick;
+} LogEntry; // logEntry holds data from AllData that is to be logged
+
+
+/* Constants -----------------------------------------------------------------*/
+static const int32_t SLOW_LOG_DATA_PERIOD_ms = 700; // logging period for slow routine
+static const int32_t FAST_LOG_DATA_PERIOD_ms = 200; // logging period for fast routine
```
zelongzhang, line 61: If we're using UPPERCASE_UNDERSCORE, the ms should also be uppercased?

Src/LogData.c
zelongzhang: I think waitUntilEEPROMReady should be good. checkEEPROM sounds like it should return a bool but it doesn't.

Src/LogData.c
zelongzhang: Careful though. I'm not too sure exactly how tight c typing is but writeToEEPROM is expecting a uint8_t*. Not sure if it matters here though.

Src/LogData.c
Matt4890: Yeah, I'd say so. If any units require lowercase letters (e.g. Bytes vs bits), we could just spell it out or use an abbreviation

Src/LogData.c
Matt4890: I think C won't care. This is where you'd use a `static_cast<uint8_t>(...)` in C++. Worth testing though.

Src/LogData.c
```diff
@@ -10,342 +19,241 @@
...
+    writeToEEPROM((char*)givenLog, LOG_ENTRY_SIZE, memAddress);
 }
 
-void readAllDataFromEEPROM(uint16_t addr, AllData* rtn, uint16_t memAddress, uint16_t timeout)
+/**
+ * @brief reads LogEntry-sized char array from EEPROM and updates the log entry
+ * @param memAddress, the address at which the EEPROM is instructed to start reading
+ * @param givenLog, pointer to a log initialized at the start of task
+ */
+void readLogEntryToEEPROM(uint16_t memAddress, LogEntry* givenLog)
```
AgamAulakh: read from ;((((

AgamAulakh: not using dataRead, instead going with this: readFromEEPROM((uint_8*)givenLog, sizeof(givenLog), memAddress); (??? let me know haha)

Src/LogData.c
```diff
@@ -9,343 +18,251 @@
...
-void buildLogEntry(AllData* data, char* buffer)
+/**
+ * @brief this method initializes the log entry struct
+ * @param givenLog, pointer to a log created at the start of the task
+ */
+void initializeLogEntry(LogEntry* givenLog)
+{
+    memset(&LogEntry, -1, LOG_ENTRY_SIZE);
+    theLogEntry.currentFlightPhase = getCurrentFlightPhase();
+    theLogEntry.tick = osKernelSysTick();
```
zelongzhang: should be givenLog instead of theLogEntry.

Src/LogData.c
```diff
@@ -9,343 +18,251 @@
...
+}
+/**
```
zelongzhang, line 176: New line before this.

zelongzhang (changes requested):

Matt4890 (approved):

PR #34, opened 2021-09-11
CubeMX Rework

Matt4890:
Removes all of the git conflicts within the `AvionicsSoftware.ioc` file, and updates the surviving configuration to be inline with the schematic (pre SPI flash changes).

TODO:
- Generate Code (Will move to STM32CubeIDE at the same time due to deprecation. Different branch & pull request)
- Figure out what `EN_7V5` and `PMB_GPIO_1` are


zelongzhang:
3.	What are EN_7V5 and PMB_GPIO1?
What are these signals coming from the SSW-106-01-F-S connector? 

"_They are both just GPIO for the power management board. EN_7V5 is the enable GPIO for the 7.5V power source which will be used to trigger the parachute ejection (mapped in Q2). PMB_GPIO1 is a secondary GPIO for future  use, does not have a purpose rn._"

I believe EN_7V5 used to be Main_Parachute by this description, not sure though. Thoughts?

Matt4890:
Aha, I was wondering where that went. I'll confirm with Randy. He said we should probably have some kind of an integration test next weekend (technically anything flight-critical should be done by the 25th), so whatever it is should be implemented or at least usable by next Saturday.

PR #41, opened 2022-05-21
IDE Swap

Matt4890:
Moves away from using both SystemWorkbench and STM32CubeMX as 2 separate programs, and now uses just STM32CubeIDE, which is essentially a combination of both of them. This is also a shift away from the previous monolithic Makefile (which gave us countless issues with onboarding and just trying to get everyone's install working properly), and now each makefile is generated by the IDE for each respective build.


Matt4890:
Worked for Kevin and Chris before rebasing onto master, and works for Agam and I now. Considering this enough testing.

PR #42, opened 2022-05-27
Clang Format

Matt4890:
Integrates clang-format in order to standardize code formatting. Includes a script to run, which formats only `.c` and `.h` files we have written, as well as `main.c`. **CLANG-FORMAT MUST BE INSTALLED LOCALLY! DOWNLOAD AND INSTALL LLVM: https://llvm.org/builds/**

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,182 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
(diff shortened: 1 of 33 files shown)


Matt4890:
Compile issues, don't merge until fixed!

PR #43, opened 2022-05-28
Solid Flight Phases and Global Header

zelongzhang:
Implements new solid flight phases according to new state diagrams. 

Convert previous tasks, MonitorForEmergencyShutoff and AbortPhase into gsListenerTask and flightPhaseTask.

Comments out hybrid specific and unused tasks (i.e ParachuteControl, EngineControl).

Adds global definition for rocket type and peripherals in new Global.h file, includes contents of old data.h file.

### DIFF
--- a/Core/Src/main.c
+++ b/Core/Src/main.c
@@ -86,6 +86,10 @@ static osThreadId transmitDataTaskHandle;
 // Debug thread
 static osThreadId debugTaskHandle;
 
+// Flight phase threads
+static osThreadId flightPhaseTaskHandle;
+static osThreadId gsListenTaskHandle;
+static osThreadId debugTaskHandle;
 // Flight phases variables
 osMutexId flightPhaseMutex;
 xQueueHandle flightPhaseQueue; //queue to be passed to flight phase thread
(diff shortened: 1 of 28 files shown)


Core/Src/main.c
```diff
@@ -361,14 +314,24 @@ int main(void)
...
     osThreadDef(
-        abortPhaseThread,
-        abortPhaseTask,
+        flightPhaseThread,
+		flightPhaseTask,
+        osPriorityHigh,
+        1,
+        configMINIMAL_STACK_SIZE * 3
+    );
+    logDataTaskHandle =
```
AgamAulakh: should this be assigned to the logDataTaskHandle?

Core/Src/main.c
```diff
@@ -361,14 +314,24 @@ int main(void)
...
+    osThreadDef(
+        gsListenThread,
+		gsListenerTask,
         osPriorityHigh,
         1,
         configMINIMAL_STACK_SIZE
     );
-    abortPhaseTaskHandle =
-        osThreadCreate(osThread(abortPhaseThread), NULL);
+    logDataTaskHandle =
```
AgamAulakh: !!

Core/Src/main.c
zelongzhang: changed to flightPhaseTaskHandle.

Core/Src/main.c
zelongzhang: changed to gsListenTaskHandle.

PR #49, opened 2022-10-14
Shivam: Timers Wrapper Code Review

### LINKED ISSUE #48: [DMB] Implement RTOS Wrapper for Timer
Intro task for DMB work, implement the FreeRTOS C++ wrapper for RTOS timers. 

Reportee: @cjchanx 

**Implementing and testing:**
Components/Core/Timer.cpp
Components/Core/Timer.hpp

**Desired Output**
- Timer object implemented and tested that supports the following:
 - Polling timers (it will update in the background, and we only care about if it is active when we poll it (for instance, calling tmr->GetState()) : please make an enum / rudimentary state diagram for the timer if possible.
- ```Timer.cpp/hpp``` is properly documented and follows coding standards

**Resources**
https://www.freertos.org/FreeRTOS-Software-Timer-API-Functions.html

**Future Activity - Will be new card once this card is reviewed**
- Expand timer into an "interrupt timer" which will automatically call a passed in callback.

[Linked Trello Card](https://trello.com/c/Dg74EmB4/58-dmb-implement-rtos-wrapper-for-timer)

cjchanx:
Leave as draft pull request please, in the future open a pull request to request for review (but we haven't gone through this yet so I've done it this way for now 😄)


Components/Core/Inc/Timer.hpp
```diff
@@ -15,6 +15,13 @@ constexpr uint32_t DEFAULT_TIMER_COMMAND_WAIT_PERIOD = MS_TO_TICKS(15); // Defau
 
 #define DEFAULT_TIMER_PERIOD (MS_TO_TICKS(1000)) // 1s
 
+enum CurrentState
+			{
+				UNINITIALIZED=0,
+				COUNTING,
+				PAUSED,
+				COMPLETE
+			};
```
cjchanx: Fix spacing

Components/Core/Inc/Timer.hpp
```diff
@@ -39,7 +49,11 @@ class Timer
 
 protected:
 	TimerHandle_t rtTimerHandle;
+	CurrentState timerState;
+	TickType_t remainingTime;
```
cjchanx: If RTOS is keeping track of this, we shouldn't have to keep track of this. But if it makes it more efficient, keep it, just remember it won't always be consistent (you must update this every time you do any operation with it).

Also, please make this a uint32_t with milliseconds and don't use TickType_t. Append the type of the variable in somewhat obscure things like this, eg remainingTimeMs

I recommend writing a GetRemainingTime() function

Components/Core/Timer.cpp
```diff
@@ -12,6 +12,11 @@
 */
 void empty_callback(TimerHandle_t rtTimerHandle) {};
 
+//CurrentState Timer::completed()
+//{
+//	timerState = COMPLETE;
+//}
+
```
cjchanx: Don't keep unused code, unless you plan to use it

Components/Core/Inc/Timer.hpp
```diff
@@ -28,7 +35,10 @@ class Timer
 	Timer();
 
 	bool ChangePeriod(const uint32_t period);
-
+	bool StartTimer();
+	bool StopTimer();
+	CurrentState GetState(TimerHandle_t xTimer);
+	CurrentState completed(TimerHandle_t xTimer);
```
cjchanx: Shouldn't need this, if we do then capitalize it.

If it's meant to be used as a callback, it'll have to be static.

Components/Core/Inc/Timer.hpp
```diff
@@ -28,7 +35,10 @@ class Timer
 	Timer();
 
 	bool ChangePeriod(const uint32_t period);
-
+	bool StartTimer();
+	bool StopTimer();
+	CurrentState GetState(TimerHandle_t xTimer);
```
cjchanx: Shouldn't have to pass in timer handle

Components/Core/Inc/Timer.hpp
```diff
@@ -39,7 +49,11 @@ class Timer
...
+	CurrentState timerState;
+	TickType_t remainingTime;
 
 };
 
-#endif /* AVIONICS_INCLUDE_SOAR_CORE_COMMAND_H */
\ No newline at end of file
+
+
+#endif /* AVIONICS_INCLUDE_SOAR_CORE_COMMAND_H */
```
cjchanx: I know I put this here, but please change to AVIONICS_INCLUDE_SOAR_CORE_TIMER_H

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
 	// We make a timer named "Timer" with a callback function that does nothing, Autoreload false, and the default period of 1s.
 	// The timer ID is specified as (void *)this to provide a unique ID for each timer object - however this is not necessary for polling timers.
 	// The timer is created in the dormant state.
-	rtTimerHandle = xTimerCreate("Timer", DEFAULT_TIMER_PERIOD, pdFALSE, (void *)this, empty_callback);
+	assert (rtTimerHandle = xTimerCreate("Timer", DEFAULT_TIMER_PERIOD, pdFALSE, (void *)this, empty_callback));
```
cjchanx: Assignment shouldn't be in an assert, hard to read. Also, we don't use system assert.

Assert rtTimerHandle **after** assignment, and use SOAR_ASSERT(cond,msg)

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
 /**
  * @brief Changes this timer object's RTOS timer period, returns true on success, returns false on failure (timer command queue full)
  */
 bool Timer::ChangePeriod(const uint32_t period)
 {
-	if (xTimerChangePeriod(rtTimerHandle, period / portTICK_RATE_MS, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdTRUE)
+	if (xTimerChangePeriod(rtTimerHandle, period / portTICK_RATE_MS, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdTRUE) {
+		timerState = COUNTING;
 		return true;
+	}
```
cjchanx: Change period/portTICK to MS_TO_TICKS

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
 	// The timer is created in the dormant state.
-	rtTimerHandle = xTimerCreate("Timer", DEFAULT_TIMER_PERIOD, pdFALSE, (void *)this, empty_callback);
+	assert (rtTimerHandle = xTimerCreate("Timer", DEFAULT_TIMER_PERIOD, pdFALSE, (void *)this, empty_callback));
+	timerState = UNINITIALIZED;
 }
 
 /**
  * @brief Changes this timer object's RTOS timer period, returns true on success, returns false on failure (timer command queue full)
  */
 bool Timer::ChangePeriod(const uint32_t period)
```
cjchanx: Please make two different functions, this one should be something like **ChangePeriodAndStart(const uint32_t period_ms)**

And the other shouldn't start the timer (or immedietly stop it) just called ChangePeriod

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
 {
-	if (xTimerChangePeriod(rtTimerHandle, period / portTICK_RATE_MS, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdTRUE)
+	if (xTimerChangePeriod(rtTimerHandle, period / portTICK_RATE_MS, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdTRUE) {
+		timerState = COUNTING;
 		return true;
+	}
 	return false;
 }
+
+
```
cjchanx: Need comments describing function

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
+bool Timer::StartTimer()
+{
+	if (xTimerStart(rtTimerHandle, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdPASS) {
+		timerState = COUNTING;
+		return true;
+	}
+	return false;
```
cjchanx: This should be start/pause, it shouldn't reset to the DEFAULT_WAIT_PERIOD it should continue on from the last period if time remained.

I recommend doing different things based on the timer state too, if we're already COUNTING then you probably don't want to do anything, for example. If we're COMPLETE we probably don't want to do anything until the remaining time is non-zero (we've changed the period).

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
+}
+
+bool Timer::StopTimer()
+{
+	if (xTimerStart(rtTimerHandle, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdPASS) {
```
cjchanx: Why is this xTimerStart

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
+		timerState = PAUSED;
+		return true;
+	}
+	return false;
+}
+
+CurrentState Timer::completed(TimerHandle_t xTimer)
```
cjchanx: Function should be capitalized, need comments. Don't pass in timerhandle (unless callback)

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
+
+bool Timer::StartTimer()
+{
+	if (xTimerStart(rtTimerHandle, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdPASS) {
+		timerState = COUNTING;
+		return true;
+	}
+	return false;
+}
+
```
cjchanx, line 102: Docs

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
+	if (xTimerStart(rtTimerHandle, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdPASS) {
+		timerState = PAUSED;
+		return true;
+	}
+	return false;
+}
+
+CurrentState Timer::completed(TimerHandle_t xTimer)
+{
+	if ((remainingTime = TICKS_TO_MS(xTimerGetExpiryTime(rtTimerHandle) - xTaskGetTickCount()) == 0)) {
```
cjchanx: Don't do assignment in an if statement like this, unclear.

It's probably better to derive this based on state. Not entirely sure why you need a function called completed (definitely shouldn't be a public function even if we need it) -- unless you meant this to be a callback

Make a function called GetTimeRemaining with that whole expirytime - tickcount statement.

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
+		return COMPLETE;
+	}
+}
+
+CurrentState Timer::GetState(TimerHandle_t xTimer)
```
cjchanx: Shouldn't have to pass this in

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
+{
+	if (xTimerIsTimerActive(rtTimerHandle) != pdFALSE)
+		return COUNTING;
+	return (completed(rtTimerHandle));
+	return timerState;
```
cjchanx: Inaccessible code. Once you've made 'completed' just a GetRemainingTime() function, do something like if(GetRemainingTime() == 0) return COMPLETE;

Also remember some checks are only needed if you are in certain states, you are storing internal state for a reason.

For example, you don't need to check Time Remaining if you are already in UNITIALIZED, COMPLETE or PAUSED, in fact you (probably) only need to check this when it is COUNTING.

I would probably make this a switch (or a set of elifs). Also, please remember to keep the internal state consistent, it's better to set the internal state variable (timerState) then return that at the end, or at least set it before returning. These timers will be used more than once.

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
+}
+
+
+
```
cjchanx, line 217: Please make sure the code compiles

You may have to add a timer instance inside run_main to see if the compiler likes the code or not

Components/Core/Timer.cpp
```diff
@@ -20,15 +25,55 @@ Timer::Timer()
...
+	return false;
+}
+
+CurrentState Timer::completed(TimerHandle_t xTimer)
+{
+	if ((remainingTime = TICKS_TO_MS(xTimerGetExpiryTime(rtTimerHandle) - xTaskGetTickCount()) == 0)) {
+		return COMPLETE;
+	}
+}
+
```
cjchanx, line 182: Docs

cjchanx (commented):
@shivamdesai04 Reviewed, feel free to ask me questions or respond to the comments directly 😃 

By the way, for code reviews we generally leave a comment regarding fixes and only allow the reviewer to do "Resolve conversation", main exception is if its trivial (eg. the comment is 'docs' and you added docs)

Components/Core/Inc/Timer.hpp
shivamdesai04: Fixed spacing

Components/Core/Timer.cpp
shivamdesai04: was not going to use that code so I removed it

Components/Core/Inc/Timer.hpp
shivamdesai04: Did you mean we don't need the completed function or all of them?
In the case of the completed function, I altered the code so I am using states instead of the completed function.

cjchanx: Just the completed function

Components/Core/Timer.cpp
shivamdesai04: What should the ChangePeriod function contain? Invoking xTimerChangePeriod automatically starts the timer so from what I understand we should convert ms to ticks in the change period function.

cjchanx: Whatever is in ChangePeriodAndStart, just stop it immediately after (and set the internal state to PAUSED instead). You can actually just call the StopTimer() function from within this function.

Components/Core/Timer.cpp
shivamdesai04: I should have fixed that up now by just returning the timer state in the function

Components/Core/Inc/Timer.hpp
```diff
@@ -15,6 +15,13 @@ constexpr uint32_t DEFAULT_TIMER_COMMAND_WAIT_PERIOD = MS_TO_TICKS(15); // Defau
 
 #define DEFAULT_TIMER_PERIOD (MS_TO_TICKS(1000)) // 1s
 
+enum CurrentState {
```
cjchanx: Please rename enum to TimerState or TimerStatus, this hpp will be in many files with state.

Components/Core/Timer.cpp
```diff
@@ -21,14 +21,63 @@ Timer::Timer()
...
 	return false;
 }
+
+
+TimerState Timer::GetState()
+{
+	if (xTimerIsTimerActive(rtTimerHandle) != pdFALSE){
+		timerState = COUNTING;
+	}
+	else if ((xTimerIsTimerActive(rtTimerHandle) == pdFALSE) && (timerState == COUNTING)) {
```
cjchanx: Swap the order of these conditions

Components/Core/Timer.cpp
```diff
@@ -21,14 +21,63 @@ Timer::Timer()
...
+		timerState = PAUSED;
 		return true;
+	}
 	return false;
 }
+
+
+TimerState Timer::GetState()
+{
+	if (xTimerIsTimerActive(rtTimerHandle) != pdFALSE){
```
cjchanx: Recommend checking only doing operations on if(timerState == COUNTING) then check is active, every other one can basically just directly return state.

Components/Core/Timer.cpp
```diff
@@ -21,14 +21,63 @@ Timer::Timer()
...
+	return false;
+}
+
+/**
+ * @brief Stops the timer, returns true on success, returns false on failure
+*/
+bool Timer::StopTimer()
 {
-	if (xTimerChangePeriod(rtTimerHandle, period / portTICK_RATE_MS, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdTRUE)
+	if ((timerState == COMPLETE) || (timerState == UNINITIALIZED) || (timerState == PAUSED)) {
```
cjchanx: Can make this != COUNTING

Components/Core/Timer.cpp
```diff
@@ -21,14 +21,63 @@ Timer::Timer()
...
+	}
+	return false;
+}
+
+/**
+ * @brief Starts the timer, returns true on success, returns false on failure
+*/
+bool Timer::StartTimer()
+{
+	if ((timerState == COMPLETE) || (timerState == COUNTING)) {
```
cjchanx, line 90: This is purely coding style, but I would make this a flat structure.

Eg. make the first statement on it's own and make the else if an if statement (since the first statement returns anyway)

Components/Core/Inc/Timer.hpp
```diff
@@ -27,8 +34,10 @@ class Timer
 public:
 	Timer();
```
cjchanx: Add a destructor ~Timer(); that calls xTimerDelete

Components/Core/Inc/Timer.hpp
```diff
@@ -27,8 +34,10 @@ class Timer
...
 
-	bool ChangePeriod(const uint32_t period);
-
+	bool ChangePeriod(const uint32_t period_ms);
+	bool StartTimer();
+	bool StopTimer();
+	TimerState GetState();
 	// WORK-IN-PROGRESS
```
cjchanx: Add reset function, and a GetPeriod function, these are all useful utilities that could be useful.

cjchanx: Also make a GetRemainingTime() function

cjchanx: And SetAutoReload and GetAutoReload

Components/Core/Timer.cpp
shivamdesai04: Done

Components/Core/Inc/Timer.hpp
shivamdesai04: Done

Components/Core/Timer.cpp
shivamdesai04: DONE

Components/Core/Timer.cpp
shivamdesai04: Done

Components/Core/Inc/Timer.hpp
```diff
@@ -27,8 +34,10 @@ class Timer
 public:
 	Timer();
 
-	bool ChangePeriod(const uint32_t period);
-
+	bool ChangePeriod(const uint32_t period_ms);
+	bool StartTimer();
```
cjchanx: Rename these to Start() and Stop()

cjchanx: It's implied that it's a timer operation

Components/Core/Timer.cpp
```diff
@@ -21,14 +21,63 @@ Timer::Timer()
...
+
+
 /**
  * @brief Changes this timer object's RTOS timer period, returns true on success, returns false on failure (timer command queue full)
  */
-bool Timer::ChangePeriod(const uint32_t period)
+bool Timer::ChangePeriod(const uint32_t period_ms)
+{
+	if (xTimerChangePeriod(rtTimerHandle, MS_TO_TICKS(period_ms), DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdTRUE) {
+		StopTimer();
```
cjchanx: Recommend making this directly call xTimerStop, and also check if that returns pdTRUE

Components/Core/Timer.cpp
shivamdesai04: Done

Components/Core/Timer.cpp
```diff
@@ -21,14 +22,179 @@ Timer::Timer()
...
+}
+
+/**
+ * @brief Default de-constructor makes a timer that can only be polled for state
+*/
+
+Timer::~Timer()
+{
+	xTimerDelete(rtTimerHandle,(DEFAULT_TIMER_COMMAND_WAIT_PERIOD*2));
+	SOAR_ASSERT(rtTimerHandle, "Error Occurred, Timer could not be deleted");
```
cjchanx: 1. Please change this to check the output of xTimerDelete instead of rtTimerHandle, this doesn't check for much other than rtTimerHandle != null.
2. Please change to an if statement and SOAR_PRINT instead, it isn't so severe that we need to reset the system over a timer not getting deleted.

Components/Core/Timer.cpp
```diff
@@ -21,14 +22,179 @@ Timer::Timer()
...
  * @brief Changes this timer object's RTOS timer period, returns true on success, returns false on failure (timer command queue full)
  */
-bool Timer::ChangePeriod(const uint32_t period)
+bool Timer::ChangePeriod(const uint32_t period_ms)
+{
+	if (xTimerChangePeriod(rtTimerHandle, MS_TO_TICKS(period_ms), DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdTRUE) {
+		if (xTimerStop(rtTimerHandle, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdPASS) {
+			timerState = PAUSED;
+			return true;
+		}
```
cjchanx, line 68: add an else here to account for if we failed xTimerStop (for whatever reason)

cjchanx:
New Pull Request:
https://github.com/StudentOrganisationForAerospaceResearch/AvionicsSoftware/pull/58

PR #52, opened 2022-10-19
Andromeda v4.1.1 Implement DebugTask

cjchanx:
- Had to unwrap HAL handles because those are externed by stm32f4xx_it.c 😭 (there is a workaround of doing the same thing we do to main.cpp, but we should wait until we have a build script implemented to deal with that automatically)
- Implemented DebugTask, currently very similar to old DebugTask using UART interrupts (Chris/dma branch has attempt at implementing DMA but it'll have some more complication to it before getting it to work...)
- Fixed GPIO.hpp not being able to be included properly in multiple files
- Add some useful utils
- Add useful seconds count to FlightTask (note this isn't necessarily how this should be done, we (maybe) should be using the RTC if possible. But if the HID task has a interval of 10x (wakes 10 times each second) 100ms or heartbeat task, then might as well use that to track it too.

DebugTask Based on code from AvionicsSoftware Legacy:
https://github.com/StudentOrganisationForAerospaceResearch/AvionicsSoftware/blob/Andromeda_V3.31_Legacy/Core/Src/Debug.c
https://github.com/StudentOrganisationForAerospaceResearch/AvionicsSoftware/blob/Andromeda_V3.31_Legacy/Core/Src/Debug.h
https://github.com/StudentOrganisationForAerospaceResearch/AvionicsSoftware/blob/Andromeda_V3.31_Legacy/Core/Src/main.c


Components/Core/Inc/Queue.hpp
```diff
@@ -34,6 +34,8 @@ class Queue{
 
 	//Functions
 	bool Send(Command& command);
+	bool SendFromISR(Command& command);
+
```
cjchanx, line 38: Make a function to automatically make a command inside the Queue, would simplify code for trivial commands.

eg. SendBasicCommand(GLOB, TASK = 0)

Components/SoarDebug/DebugTask.cpp
```diff
@@ -20,7 +24,32 @@ constexpr uint8_t DEBUG_TASK_PERIOD = 100;
 
 /* Prototypes ----------------------------------------------------------------*/
 
+/* HAL Callbacks ----------------------------------------------------------------*/
+/**
+ * @brief HAL Callback for DMA/Interrupt Complete
+ *
+ * TODO: This should eventually be in DMAController or main_avionics.cpp depending on how many tasks use DMA vs Interrupt vs Polling
+ */
```
cjchanx, line 33: Or UARTTask

Components/SoarDebug/DebugTask.cpp
```diff
@@ -39,15 +68,97 @@ void DebugTask::InitTask()
...
+ * @brief Receive data to the buffer
+ * @return Whether the debugBuffer is ready or not
+ */
+void DebugTask::InterruptRxData()
+{
+	// If we already have an unprocessed debug message, ignore this byte
+	if (!isDebugMsgReady) {
+		// Check byte for end of message - note if using termite you must turn on append CR
+		if (debugRxChar == '\r' || debugRxChar == '\0' || debugMsgIdx == DEBUG_RX_BUFFER_SZ_BYTES) {
+			// Null terminate and process
```
cjchanx, line 152: Probably don't need to check '/0', can assume carriage return is terminator

Components/SoarDebug/DebugTask.cpp
```diff
@@ -39,15 +68,97 @@ void DebugTask::InitTask()
...
+			debugBuffer[debugMsgIdx++] = '\0';
+
+			// Notify the debug task
+			Command cm(DATA_COMMAND, EVENT_DEBUG_RX_COMPLETE);
+			qEvtQueue->SendFromISR(cm);
```
cjchanx: If this fails, we need to clear the buffer and re-arm the interrupt (otherwise we'll stall, even though this should really never happen)

Components/Core/Inc/Queue.hpp
cjchanx: Will be implemented later.

AgamAulakh (approved):

PR #57, opened 2022-12-10
Andromeda v4.2.0 (Rocket State Machine and Rocket State Commands)

cjchanx:
Rocket State Machine, State Commands Enum, Debug with Parameters

RocketSM object handles the state machine, each state is an inherited class from a base rocket state, where the RocketSM's current state pointer stores a pointer to the current state.

Full-Cycle Run Logs:
```
-- SOAR AVIONICS --
System Reset Reason: [TODO]
Current System Heap Use: 10904 Bytes
Lowest Ever Heap Size: 10856 Bytes

Rocket State Machine Started in [ Abort ] state
rsc 11
ROCKET STATE TRANSITION [ Abort ] --> [ Pre-launch ]
rsc 7
ROCKET STATE TRANSITION [ Pre-launch ] --> [ Fill ]
rsc 11
ROCKET STATE TRANSITION [ Fill ] --> [ Pre-launch ]
rsc 7
ROCKET STATE TRANSITION [ Pre-launch ] --> [ Fill ]
rsc 8
rsc 9
rsc 10
ROCKET STATE TRANSITION [ Fill ] --> [ Arm ]
rsc 11
rsc 7
ROCKET STATE TRANSITION [ Arm ] --> [ Fill ]
rsc 8
rsc 9
rsc 10
ROCKET STATE TRANSITION [ Fill ] --> [ Arm ]
rsc 14
ROCKET STATE TRANSITION [ Arm ] --> [ Ignition ]
rsc 10
ROCKET STATE TRANSITION [ Ignition ] --> [ Arm ]
rsc 14
ROCKET STATE TRANSITION [ Arm ] --> [ Ignition ]
rsc 15
ROCKET STATE TRANSITION [ Ignition ] --> [ Launch ]
rsc 16
ROCKET STATE TRANSITION [ Launch ] --> [ Burn ]
rsc 17
ROCKET STATE TRANSITION [ Burn ] --> [ Coast ]
rsc 18
ROCKET STATE TRANSITION [ Coast ] --> [ Descent ]
rsc 19
ROCKET STATE TRANSITION [ Descent ] --> [ Recovery ]
rsc 1
ROCKET STATE TRANSITION [ Recovery ] --> [ Abort ]
rsc 7
rsc 11
ROCKET STATE TRANSITION [ Abort ] --> [ Pre-launch ]
rsc 1
ROCKET STATE TRANSITION [ Pre-launch ] --> [ Abort ]
```


Components/Core/Inc/Command.hpp
```diff
@@ -19,7 +19,8 @@ enum GLOBAL_COMMANDS : uint8_t
 {
 	COMMAND_NONE = 0,		// No command, packet can probably be ignored
 	TASK_SPECIFIC_COMMAND,	// Runs a task specific command when given this object
-	DATA_COMMAND			// Data command, used to send data to a task. Target is stored in taskCommand
+	DATA_COMMAND,			// Data command, used to send data to a task. Target is stored in taskCommand
+    CONTROL_ACTION,			// Control actions, used in Rocket State Machine, direct translation to RCU<->DMB Protocol
```
AgamAulakh, line 23: tab, also shouldnt there be a max to handle an invalid type of command?

Components/Core/Inc/Command.hpp
```diff
@@ -19,7 +19,8 @@ enum GLOBAL_COMMANDS : uint8_t
...
 };
 
 /* Class -----------------------------------------------------------------*/
```
AgamAulakh, line 26: I know we've talked about this before, but I am still slightly concerned about the way this manages memory. A command could either allocate its own buffer or point to a previously-allocated buffer, all the while managing the call to soar_free using a bool. I think (and I want to hear your opinion) this provides too much functionality to someone using a command obj and puts the onus on them to manage the different configurations properly. Also, if the same command is raw-copied to multiple queues (eg. state broadcast to multiple threads), there would be no way of knowing when the last reference to the heap ptr is gone.

Ideally, we use malloc and free in the same scope to be explicit, but this isn't possible in this case. Maybe we use pointers to stack memory like a typedef uint8_t[known_imu_data_buffer_sz] imu_data_packet.

Components/FlightControl/Inc/RocketSM.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+    //-- RECOVERY / TECHNICAL --
+    RS_ABORT,       // Abort sequence, vents open, MEV closed, ignitors off
+    RS_NONE         // Invalid state, must be last
+};
+
+/**
+ * @brief Rocket State Commands, all fall under the GLOBAL_COMMAND -> STATE_ACTION umbrella
+ *
+ *        All commands in this list are only activated if the state allows them.
```
AgamAulakh: "only activated if the state allows" - how so? I don't quite understand what these are for, are they the thread-specific state commands? Do we need them?

Components/FlightControl/Inc/RocketSM.hpp
```diff
@@ -0,0 +1,295 @@
...
+    //-- PRELAUNCH --
+    RSC_BEGIN_FILL, // Transition to the FILL state
+
+    //-- FILL --
+    RSC_ARM_CONFIRM_1,   // Enable first ARM confirmation flag
+    RSC_ARM_CONFIRM_2,   // Enable second ARM confirmation flag
+    RSC_ARM_ACTION,      // Transition to the ARM state
+
+    //-- ARM/IGNITION/LAUNCH/BURN --
+    RSC_MANUAL_OVERRIDE_ENABLE, // For any states with locked out capability. If this flag is set using a command the NEXT command will be allowed to directly control valves/vents
```
AgamAulakh: continued from our discussion last week: I don't think we need to program this logic. The operator is expected to know when they are performing "manual override" actions (with a helpful reminder from the front end)

Components/FlightControl/Inc/RocketSM.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+    //-- ARM --
+    RSC_POWER_TRANSITION_ONBOARD,      // Change power source to onboard
+    RSC_POWER_TRANSITION_EXTERNAL,     // Change power source to external power
+    RSC_READY_FOR_IGNITION, // Ready for ignition sequence - Transition to IGNITION state
+
+    //-- IGNITION --
+    RSC_CONFIRM_IGNITION,   // Confirm igniter actuation - Transition to LAUNCH state (MEV OPEN)
+    //* TBD - To ensure we don't get stuck unable to vent, we have override actions allowing a transition to ABORT
+    RSC_CRITICAL_IGNITION_ALLOW_ABORT, // Enable flag allowing us to abort in the ignition state
```
AgamAulakh: again, unnecessary as per our last discussion

Components/FlightControl/Inc/RocketSM.hpp
AgamAulakh: continued from our discussion last week: I don't think we need to program this logic. The operator is expected to know when they are performing "manual override" actions (with a helpful reminder from the front end)

Components/FlightControl/Inc/RocketSM.hpp
```diff
@@ -0,0 +1,295 @@
...
+    
+
+    //-- LAUNCH/BURN --
+    //* TBD - To ensure we don't get stuck unable to vent, we have override actions for LAUNCH/BURN
+    RSC_CRITICAL_MEV_CLOSE_OVERRIDE_CONFIRM, // Enable confirmation flag for MEV close override
+    RSC_CRITICAL_MEV_CLOSE_OVERRIDE_ACTION,  // Action for MEV override, requires override flag is set
+    RSC_CRITICAL_BURN_ALLOW_ABORT, // Enable flag allowing us to abort in the ignition state
+    RSC_CRITICAL_CLEAR_OVERRIDE_FLAGS,      // Clears any override flags in either IGNITION/LAUNCH/BURN
```
AgamAulakh: Maybe I'm not understanding the intention here. To me, this seems unnecessary like the last few overrides, but let's chat :)

Components/FlightControl/Inc/RocketSM.hpp
```diff
@@ -0,0 +1,295 @@
...
+    //-- PRE-IGNITION and RECOVERY --
+    RSC_ABORT,       // Transition to ABORT state - available from all states except for IGNITION/LAUNCH/BURN
+    RSC_OPEN_VENT,   // Open the vent valve
+    RSC_CLOSE_VENT,  // Close the vent valve
+    RSC_OPEN_DRAIN,  // Open the drain valve
+    RSC_CLOSE_DRAIN, // Close the drain valve
+    RSC_MEV_CLOSE,   // Forces MEV to close - ONLY supported in states where it is safe to close the MEV
+
+    //-- PRELAUNCH --
+    RSC_BEGIN_FILL, // Transition to the FILL state
```
AgamAulakh: let's follow a naming convention for the Transition-To-State commands (ie. it could literally just be RSC_TRANSITION_TO_FILL)

Components/FlightControl/Inc/RocketSM.hpp
```diff
@@ -0,0 +1,295 @@
...
+};
+
+/**
+ * @brief Base class for Rocket State Machine
+ */
+class BaseRocketState
+{
+public:
+    virtual RocketState HandleCommand(Command& cm) = 0; //Handle a command based on the current state
+    virtual RocketState OnEnter() = 0;  //Returns the state we're entering
```
AgamAulakh, line 113: really excited about this! I didn't like the idea of having an array of states, but I went through the FSM example with ETL and it looked really clean.

On another note, I'm assuming OnEnter() would be used for broadcasting state changes to threads and starting timers. I also feel like OnExit() might not be used, but its not bad to have for now

Components/FlightControl/RocketSM.cpp
```diff
@@ -0,0 +1,455 @@
...
+ */
+RocketState RocketControl::TransitionState(RocketState nextState)
+{
+    // Check if we're already in the next state (TransitionState does not allow entry into the existing state)
+    if (nextState == rs_currentState->GetStateID())
+        return rs_currentState->GetStateID();
+
+    // Check the next state is valid
+    if (nextState >= RS_NONE)
+        return rs_currentState->GetStateID();
```
AgamAulakh, line 57: also need to check if a transition to nextState is possible from rs_currentState (ie. this is where we program hard locks on transitions, for example, from fill to ignition)

Components/FlightControl/RocketSM.cpp
```diff
@@ -0,0 +1,455 @@
...
+ */
+Ignition::Ignition()
+{
+    rsStateID = RS_IGNITION;
+}
+
+/**
+ * @brief Entry to Arm state
+ * @return The state we're entering
+ */
```
AgamAulakh: clean up: remainder of file is a copy of arm

Components/FlightControl/RocketSM.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+    // Handle the command based on the current state
+    rs_currentState->HandleCommand(cm);
+}
+
+/* Base State ------------------------------------------------------------------*/
+/**
+ * @brief General handler for actions that should be supported by all rocket state machines
+ */
+RocketState BaseRocketState::HandleGeneralStateCommands(RocketControlCommands rcAction)
```
AgamAulakh: Im a bit confused with when these Handle_X_StateCommands() will be used. Let me know if this is correct: each state has a state-specific HandleCommand() function and this is where we block transitions to invalid/unreachable states. (how is this chained with RocketControl::TransitionState()?). Only prelaunch can call HandleNonIgnitionCommands (which, by the way, is probably a general task that should be implemented by the base class along with sanity checks using rsStateID). We also only call HandleGeneralStateCommands() in prelaunch, but it's called at the start along with the other command checks...?

AgamAulakh (changes requested):

Components/Core/Inc/Command.hpp
cjchanx: Invalid commands which aren't implemented aren't processed, it'd end up in the default condition for the switch.

Components/Core/Inc/Command.hpp
cjchanx: A command should never be copied, each command object should be its own instance. 'Broadcast' is handled simply by making multiple command objects each with it's own memory. A pointer to stack memory can be used in the existing command structure, that's the point of having the ability to point to a pre-allocated buffer. However, in most cases we'll be using a command owned buffer as it makes it possible to use the assumption, "**if I own the command, I own that memory**".

How should malloc and free be used in the same scope? In most cases where memory is only necessary in one scope, and needs to be freed, it should be using stack? (exception being static buffers / objects)

Could you clarify what you mean by pointer to stack memory here? Would that not be equivalent to using a pre-allocated buffer? If you mean having the 'command' actually take the entire size of storage itself that would limit the queue depths which we could use by alot.

Components/FlightControl/Inc/RocketSM.hpp
cjchanx: These rocket state commands aren't used, this is why I mentioned this is an incomplete PR.

Components/FlightControl/Inc/RocketSM.hpp
cjchanx: Sure... but I will note something, if the DMB is going to allow everything from every state anyway (with no checks, too), the state machine is effectively being completely ignored... In that case, why aren't we just programming state functionality from the frontend?

Components/FlightControl/Inc/RocketSM.hpp
cjchanx: We should be able to abort directly from ignition states? Doesn't that fundamentally go against the state diagram? 
image

Or do you mean don't allow abort whatsoever from these states?

Components/FlightControl/RocketSM.cpp
cjchanx: Hard locks on transitions are implemented by the fact that a given state's HandleCommand function does not implement the functionality to transition to a given state. It should not be necessary to implement a double lock here.

Components/FlightControl/RocketSM.cpp
cjchanx: Any given HandleCommand function **only** implements functionality that specific state is allowed to process. If a specific set of functionality can be shared between multiple states we implement that by deriving that state and calling the parent class' unique function (to avoid code duplication). HandleNonIgnitionCommands is, as it sounds, "a set of commands that can be done by any state prior to ignition eg. Abort"

This specific function was implemented based on my previous assumption that we'd have any sort of security on commands from being executed in a given state... but given that we're apparently going to allow everything from every state anyway..... this can likely be folded into a general "HandleGeneralCommands" function, with only abort being handle d in this particular function.

Components/FlightControl/Inc/RocketSM.hpp
cjchanx: Yes, most functionality of a given state (not related to commands) will be implemented inside OnEnter.

OnExit is used for things like clearing flags or general cleanup.

For example, in the Fill:: state we use two arm confirmation flags, we clear those once we exit that state.

Components/FlightControl/RocketSM.cpp
cjchanx: Yeah that's part of the work in progress lol, fixed some of the comments here: 0dc6df9313551d45e98a62cb876ceddb29588b67

Components/FlightControl/Inc/RocketSM.hpp
cjchanx: Removed here: 16db7e4ce019c4ed7ae58415f6b272136c77dd79

Components/FlightControl/Inc/RocketSM.hpp
cjchanx: This is one of the things I wanted to clarify. Do we want to allow the MEV to close in launch/burn states by just a single command as-is? I thought it was mentioned (in that rocket state meeting) that it _really_ shouldn't close during the launch/burn states.

Components/FlightControl/Inc/RocketSM.hpp
cjchanx: Added TRANSITION to the name of state change commands: 1ad467b9575ad0deb3796a3431f839adda189b27

Note that I kept a little more context in the names eg. "CONFIRM_IGNITION" instead of simply "IGNITION" for state transitions that imply a kind of process

Components/Core/Inc/Command.hpp
cjchanx: Simplified command external buffer allowance here: 70ca4598200bd3b1aef617d946752636ebf8f834

AgamAulakh (approved):

Components/FlightControl/FlightTask.cpp
```diff
@@ -31,65 +41,73 @@ void FlightTask::InitTask()
...
-		//	osDelay(BUZZER_ERR_PERIOD);
-		//	HAL_TIM_PWM_Stop(&htim2, TIM_CHANNEL_1);
-		//}
-
-		// TODO: Message beeps
-	}
+    uint32_t tempSecondCounter = 0; // TODO: Temporary counter, would normally be in HeartBeat task or HID Task, unless FlightTask is the HeartBeat task
+    GPIO::LED1::Off();
+
+    rsm_ = new RocketSM(RS_ABORT);
```
cjchanx: Should run OnEnter for whichever state it starts in. Actually maybe not, if it's already in a given state, for instance, Launch, we don't need to send the start opening command. Only if we fully reset the state machine and enter ABORT do we need to run this, perhaps leave this up to the implementation of the recovery procedure, possibly assume that entering ABORT or RECOVERY requires an initializing OnEnter that valves etc. are open.

Maybe make this a parameter in the CTOR for now (shouldRunEnterState)

PR #58, opened 2023-01-07
Andromeda v4.3.1 (C++ FreeRTOS Timer Wrapper)

shivamdesai04:


Components/Core/Inc/Timer.hpp
```diff
@@ -25,21 +35,32 @@ constexpr uint32_t DEFAULT_TIMER_COMMAND_WAIT_PERIOD = MS_TO_TICKS(15); // Defau
...
+	bool ResetTimer();
+	bool ResetTimerAndStart();
+	void SetAutoReload(bool setReloadOn);
 
+	bool GetAutoReload();
+	TimerState GetState();
+	const uint32_t GetPeriod();
+	const uint32_t GetRemainingTime();
 
+	static void CallbackFunction( TimerHandle_t xTimer );
```
cjchanx: Please rename to DefaultCallback

Components/Core/Inc/Timer.hpp
```diff
@@ -25,21 +35,32 @@ constexpr uint32_t DEFAULT_TIMER_COMMAND_WAIT_PERIOD = MS_TO_TICKS(15); // Defau
...
 
 protected:
 	TimerHandle_t rtTimerHandle;
-
+	TimerState timerState;
+	uint32_t remainingTime;
+	uint32_t remainingTimeBetweenPauses;
+	uint32_t rtosTimeRemaning();
```
cjchanx: Spelling, also please add a few lines between this function and the variables, ideally put the function at the top, also please rename the function with a capital, like this:

```
	const uint32_t GetRTOSTimeRemaining() const;

	TimerHandle_t rtTimerHandle;
	TimerState timerState;
	uint32_t remainingTime;
	uint32_t remainingTimeBetweenPauses;
```

Components/Core/Inc/Timer.hpp
```diff
@@ -25,21 +35,32 @@ constexpr uint32_t DEFAULT_TIMER_COMMAND_WAIT_PERIOD = MS_TO_TICKS(15); // Defau
...
+	bool Start();
+	bool Stop();
+	bool ResetTimer();
+	bool ResetTimerAndStart();
+	void SetAutoReload(bool setReloadOn);
 
+	bool GetAutoReload();
+	TimerState GetState();
+	const uint32_t GetPeriod();
+	const uint32_t GetRemainingTime();
```
cjchanx: Not a huge deal, but these getters could all return const and be const functions, eg:
```
const bool GetAutoReloadEnabled() const;
```

Would be nice to rename slightly ambiguous things like:
* GetAutoReload to GetAutoReloadEnabled
* GetPeriod to GetPeriodMs() -- or whatever unit it is
* GetRemainingTime to GetRemainingTimeMs() -- or whatever unit it is

cjchanx (changes requested):

Components/Core/Timer.cpp
```diff
@@ -1,34 +1,226 @@
+#include "SystemDefines.hpp"
```
cjchanx: Please move these below the file header

Components/Core/Timer.cpp
```diff
@@ -1,34 +1,226 @@
...
-#include "Timer.hpp"
 
 /**
- * @brief Empty callback function, used internally for default polling timers
-*/
-void empty_callback(TimerHandle_t rtTimerHandle) {};
-
-/**
- * @brief Default constructor makes a timer that can only be polled for state
+ * Default constructor makes a timer that can only be polled for state
```
cjchanx: This should have a @brief at the start

Components/Core/Timer.cpp
```diff
@@ -1,34 +1,226 @@
...
+	SOAR_ASSERT(rtTimerHandle, "Error Occurred, Timer not created");
+	timerState = UNINITIALIZED;
+}
+
+/**
+ * @brief Default de-constructor makes a timer that can only be polled for state
+*/
+Timer::~Timer()
+{
+	if (xTimerDelete(rtTimerHandle, DEFAULT_TIMER_COMMAND_WAIT_PERIOD*2) == pdPASS) {
```
cjchanx: If this fails, (else block) please print out something like "Warning, failed to delete timer"

Components/Core/Timer.cpp
```diff
@@ -1,34 +1,226 @@
...
+		SOAR_PRINT("Timer has been deleted \n\n");
+	}
+}
+
+/**
+ * @brief Callback function for polling timers
```
cjchanx: Change comment to mention it must be called by all timers

Components/Core/Timer.cpp
```diff
@@ -1,34 +1,226 @@
...
+*/
+void Timer::CallbackFunction(TimerHandle_t xTimer){
+	Timer* ptrTimer = (Timer*)pvTimerGetTimerID(xTimer);
+	ptrTimer->timerState = COMPLETE;
 }
 
 /**
- * @brief Changes this timer object's RTOS timer period, returns true on success, returns false on failure (timer command queue full)
+ * @brief Changes timer period, Sets timer state back to uninitialized and stops timer
```
cjchanx, line 65: All functions with a return need an @return explaining what it returns

Components/Core/Timer.cpp
```diff
@@ -1,34 +1,226 @@
...
+/**
+ * @brief Starts the timer
+*/
+bool Timer::Start()
+{
+	if ((timerState == COMPLETE) || (timerState == COUNTING)) {
+		return false;
+	}
+	else if (timerState == PAUSED) {
+		ChangePeriod(remainingTimeBetweenPauses);
```
cjchanx: If this is ChangePeriodAndStart, then we shouldn't also need to xTimerStart (Right now it will basically Change period (start), stop, then start again, it seems)

Components/Core/Timer.cpp
```diff
@@ -1,34 +1,226 @@
...
+	}
+	if (xTimerReset(rtTimerHandle, DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdPASS) {
+		timerState = COUNTING;
+		return true;
+	}
+	return false;
+}
+
+/**
+ * @brief Sets timer to auto-reload if parameter is set to true, Sets timer to one shot if parameter is set to false
```
cjchanx: Please add an @param setReloadOn something like "auto reload to on if true, auto reload off if false"

Components/Core/Timer.cpp
```diff
@@ -1,34 +1,226 @@
...
+	}
+	if (setReloadOn == false){
+		vTimerSetReloadMode(rtTimerHandle, pdFALSE);
+		//Testing purposes
+		SOAR_PRINT("Set to One Shot\n\n");
+	}
+}
+
+/**
+ * @brief Returns true if the timer is set to autoreload and false if it is set to one-shot
```
cjchanx: Not that important, only matters if we auto-generated documentation...

This brief could be changed to @return, and @brief can be something simple like Gets auto reload state

Components/Core/Timer.cpp
```diff
@@ -1,34 +1,226 @@
...
+/**
+ * @brief Returns timer state enum
+*/
+TimerState Timer::GetState()
+{
+	return timerState;
+}
+
+/**
+ * @brief Returns the timers' period
```
cjchanx: Comment should include units

Components/Core/Inc/Timer.hpp
shivamdesai04: Done!

Components/Core/Timer.cpp
shivamdesai04: wouldn't this help for testing purposes only because we would never actually see this message anyways for actual flight

Components/Core/Inc/Timer.hpp
shivamdesai04: do you mind giving the names one more check i changed some

Components/Core/Timer.cpp
cjchanx: Good point, depends on if we route traces over radio (which I want, but we'll have to see if the system can handle it)

I'd add it either way, since rtos things not deleting would be quite bad, it's more important than the "Timer deleted" message.

(when everything is running at the same time we won't always know exactly when a timer should be deleted or not, but we should know if it can't be deleted for any reason)

Components/Core/Timer.cpp
```diff
@@ -1,34 +1,226 @@
...
+
+/**
+ * @brief Returns true if the timer is set to autoreload and false if it is set to one-shot
+*/
+bool Timer::GetAutoReload()
+{
+	if ((uxTimerGetReloadMode(rtTimerHandle)) == (( UBaseType_t ) pdTRUE)) {
+		return true;
+	}
+	if ((uxTimerGetReloadMode(rtTimerHandle)) == (( UBaseType_t ) pdFALSE)) {
```
cjchanx: Can replace this second function call with an else, it can only return pdTRUE or pdFALSE.

Components/Core/Inc/Timer.hpp
cjchanx: yep, don't see changes though, maybe you haven't commit/push?

Components/Core/Timer.cpp
shivamdesai04: yep makes sense. also might help if this flight log is actually stored. it might help for debugging after flight if we had any issues

Components/Core/Inc/Timer.hpp
```diff
@@ -25,21 +35,34 @@ constexpr uint32_t DEFAULT_TIMER_COMMAND_WAIT_PERIOD = MS_TO_TICKS(15); // Defau
...
 
+	static void DefaultCallback( TimerHandle_t xTimer );
 
 protected:
-	TimerHandle_t rtTimerHandle;
+	const uint32_t GetRTOSTimeRemaining();
+
+	TimerState timerState; // Enum that holds current timer state
+	TimerHandle_t rtTimerHandle; //what
+	uint32_t remainingTime;
```
cjchanx: Not needed, need to worry about the original timer period after a start/stop/complete.

You will need to restore the `originalTimerPeriod` in these cases:
- On `ResetTimer()` / `ResetTimerAndStart()`
- Whenever the timer is complete (inside the `TimerCallback`)

Components/Core/Timer.cpp
shivamdesai04: done!

Components/Core/Timer.cpp
```diff
@@ -8,27 +8,234 @@
...
+    ptrTimer->timerState = COMPLETE;
+}
+
+/**
+ * @brief Changes timer period, Sets timer state back to uninitialized and stops timer
+ * @return Returns true if the period is successfully changed and stopped and returns false otherwise
+ */
+bool Timer::ChangePeriodMs(const uint32_t period_ms)
+{
+    if (xTimerChangePeriod(rtTimerHandle, MS_TO_TICKS(period_ms), DEFAULT_TIMER_COMMAND_WAIT_PERIOD) == pdTRUE) {
```
cjchanx, line 70: In the future, let's keep it consistent (either pdTRUE or pdPASS)

Components/Core/Inc/Timer.hpp
```diff
@@ -25,21 +35,35 @@ constexpr uint32_t DEFAULT_TIMER_COMMAND_WAIT_PERIOD = MS_TO_TICKS(15); // Defau
...
 
+    static void DefaultCallback( TimerHandle_t xTimer );
 
 protected:
-	TimerHandle_t rtTimerHandle;
+    const uint32_t GetRTOSTimeRemaining();
+
+    TimerState timerState; // Enum that holds current timer state
+    TimerHandle_t rtTimerHandle;
+    uint32_t timerPeriod = DEFAULT_TIMER_PERIOD;
```
cjchanx, line 62: Add a quick comment `// the 'true' original period on resets`

cjchanx (approved):

PR #59, opened 2023-01-28
Jesse/rocket control functionality

Jgerbrandt:


AvionicsSoftware.ioc
```diff
@@ -165,46 +193,74 @@ PA15.GPIO_Label=BUZZER_PWM
...
 PA7.Locked=true
 PA7.Mode=Full_Duplex_Master
 PA7.Signal=SPI1_MOSI
 PA8.GPIOParameters=GPIO_Label
-PA8.GPIO_Label=LAUNCH
+PA8.GPIO_Label=BAT_EN
 PA8.Locked=true
 PA8.Signal=GPIO_Output
+PA9.GPIOParameters=GPIO_Label
+PA9.GPIO_Label=RCU_RS485_USART1_TX
```
AgamAulakh: hard to comment on an IOC file, but I think we should not add "RS485" to the pin names. To the STM, these are used as UART lines and firmware doesn't care if they're converted to a specific kind of differential pair comm

AvionicsSoftware.ioc
```diff
@@ -165,46 +193,74 @@ PA15.GPIO_Label=BUZZER_PWM
...
+PA9.Locked=true
+PA9.Mode=Asynchronous
+PA9.Signal=USART1_TX
+PB1.GPIOParameters=GPIO_Label
+PB1.GPIO_Label=PS1_ADC1_IN9
```
AgamAulakh: nitpick: I don't think we should include the specific pin type in the names. When we use the user-defined names in our code, we should be blind to, for example, the actual uart line/adc channel being used. I suggest we remove the actual pin name that's kept in the middle (like changing GPS_UART4_RX to just GPS_RX). I'm definitely open to discussion on this point!

AgamAulakh: Although the schematic has this as "PS1," I think we should just call it AV_Pressure or something along those lines

AvionicsSoftware.ioc
```diff
@@ -147,13 +164,24 @@ NVIC.UART5_IRQn=true\:5\:0\:false\:false\:true\:true\:true\:true\:true
...
 NVIC.UsageFault_IRQn=true\:0\:0\:false\:false\:true\:false\:false\:false\:false
 PA0-WKUP.GPIOParameters=GPIO_Label
-PA0-WKUP.GPIO_Label=GPS_UART_TX
+PA0-WKUP.GPIO_Label=GPS_UART4_TX
+PA0-WKUP.Locked=true
 PA0-WKUP.Mode=Asynchronous
 PA0-WKUP.Signal=UART4_TX
 PA1.GPIOParameters=GPIO_Label
-PA1.GPIO_Label=GPS_UART_RX
+PA1.GPIO_Label=GPS_UART4_RX
```
AgamAulakh: Note to Chris: GPS uart has been configured to use DMA (GPS --> mcu memory)

AvionicsSoftware.ioc
```diff
@@ -165,46 +193,74 @@ PA15.GPIO_Label=BUZZER_PWM
...
 PA6.Mode=Full_Duplex_Master
 PA6.Signal=SPI1_MISO
 PA7.GPIOParameters=GPIO_Label
-PA7.GPIO_Label=IMU_SPI_MOSI
+PA7.GPIO_Label=IMU_SPI1_MOSI
 PA7.Locked=true
 PA7.Mode=Full_Duplex_Master
 PA7.Signal=SPI1_MOSI
 PA8.GPIOParameters=GPIO_Label
-PA8.GPIO_Label=LAUNCH
```
AgamAulakh: We should configure the gpio _input_ pins (LAUNCH and DMB_ABORT) to be _pulled down_

AgamAulakh: AUX_2 should be pull up. Though we don't currently use this pin, in the past we have used the auxiliary pins for debugging by holding them to ground.

Components/FlightControl/RocketSM.cpp
```diff
@@ -155,24 +153,23 @@ RocketState PreLaunch::HandleNonIgnitionCommands(RocketControlCommands rcAction,
         // Transition to abort state
         return RS_ABORT;
```
AgamAulakh, line 190: This doesn't have to be included in this PR, but we should think about making the LAUNCH and DMB_ABORT input pins trigger an interrupt in order to immediately queue the right state transition command.

Components/FlightControl/RocketSM.cpp
```diff
@@ -348,10 +345,12 @@ RocketState Arm::HandleCommand(Command& cm)
...
+            PowerSelect::UmbilicalPower();
+            //TODO: we should check to make sure umbilical power is available before doing so
             break;
         case RSC_POWER_TRANSITION_ONBOARD:
-            //TODO: Transition to onboard power
+            PowerSelect::InternalPower();
+            //TODO: we should check to make sure internal power is available before doing so
             break;
         case RSC_GOTO_IGNITION:
             // Transition to ready for ignition state
```
AgamAulakh, line 400: Got some more clarification on the ignition state transitions from Mo:
As soon as we enter the ignition state, we should start a timer for the igniter delay. If the timer expires before the launch command is received, we should go into Arm. If the launch command is received, we should wait until the timer expires before entering launch and opening the MEV

AgamAulakh (changes requested):
Great job! Most of these comments are discussion points

cjchanx (changes requested):
Last commit has modified alot of files unintended, likely due to autocrlf and Macs leaving a .DS_Store file.

Run `git config --global core.autocrlf true` on any Windows desktops to keep consistent line endings.

Components/FlightControl/Inc/RocketSM.hpp
```diff
@@ -53,6 +53,8 @@ enum RocketState
  */
 enum RocketControlCommands
 {
+    //open vent and drain at the same time
+
```
cjchanx: Remove these two lines

Core/.DS_Store
cjchanx, line 1: Remove :)

Components/FlightControl/TelemetryTask.cpp
```diff
@@ -96,3 +96,23 @@ void TelemetryTask::RunLogSequence()
     //TODO: Commented out for now, until merged with the flight task changes
```
cjchanx, line 97: Call this function in RunLogSequence

cjchanx (approved):

PR #62, opened 2023-03-15
Marcelo/flash

VioletLiGonzales:


Components/Flash/FlashTask.cpp
```diff
@@ -0,0 +1,59 @@
...
+ * @param pvParams RTOS Passed void parameters, contains a pointer to the object instance, should not be used
+ */
+void FlashTask::Run(void * pvParams)
+{
+    st_ = new SystemStorage();
+
+    while (1) {
+        //Process any commands, in non-blocking mode
+        Command cm;
+        bool res = qEvtQueue->Receive(cm);
```
cjchanx: Let's change this to use `ReceiveWait(cm)` and change the comment to "in blocking mode" since this task isn't using periodic commands right now

Components/Flash/Inc/SystemStorage.hpp
```diff
@@ -0,0 +1,56 @@
+/**
+ ******************************************************************************
+ * File Name          : FlashTask.hpp
```
cjchanx: Change the file name here

Components/Flash/Inc/SystemStorage.hpp
```diff
@@ -0,0 +1,56 @@
...
+public:
+    SystemStorage();
+
+    void HandleCommand(Command& cm);
+
+    void WriteStateToFlash();
+    bool ReadStateFromFlash();
+    void WriteDataToFlash(uint8_t* data, uint16_t size);
+    bool ReadDataFromFlash();
+
```
cjchanx, line 50: Add a quick getter

```
StateInformation GetStateData() { return rs_currentInformation; }
```

Components/Flash/SystemStorage.cpp
```diff
@@ -0,0 +1,206 @@
...
+                SOAR_PRINT("state written to flash\n");
+            }
+            else if(cm.GetTaskCommand() == DUMP_FLASH_DATA) 
+            {
+                ReadDataFromFlash();
+            }
+            else if(cm.GetTaskCommand() == ERASE_ALL_FLASH)
+            {
+                W25qxx_EraseChip();
+            }
```
cjchanx, line 230: add a `break;` to the end of each case statement

VioletLiGonzales: Done

PR #69, opened 2023-06-04
Andromeda v4.6.1 (Human Device Interface Task)

graceejangg:
APB2 prescaler changed 2 in the ioc clock configuration


Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,182 @@
...
+    {RS_FILL,{3, 1000}},
+    {RS_ARM, {4, 1000}},
+    {RS_IGNITION, {5, 1000}},
+    {RS_LAUNCH, {6, 1000}},
+    {RS_BURN, {7, 1000}},
+    {RS_COAST, {8, 1000}},
+    {RS_DESCENT,{9, 1000}},
+    {RS_RECOVERY, {10, 1000}},
+    {RS_ABORT,{1, 1000}}
+};
```
cjchanx: Try to change to etl map

```
#include "etl/map.h"

std::map -> etl::map<RocketState,BLINK, map_max_size>
```

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+
+
+/**
+* @brief Initialize the HDITask
+*/
+void HDITask::InitTask()
+{
+    // Make sure the task is not already initialized
+    SOAR_ASSERT(rtTaskHandle == nullptr, "Cannot initialize flight task twice");
```
cjchanx: flight to HDI, fix extra spacing

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,182 @@
...
+{
+    //TODO: Since this task will stall for a few milliseconds, we may need a way to eat the whole queue (combine similar eg. REQUEST commands and eat to WDG command etc)
+    //TODO: Maybe a HandleEvtQueue instead that takes in the whole queue and eats the whole thing in order of non-blocking to blocking
+
+
+
+    //Switch for the GLOBAL_COMMAND
+    switch (cm.GetCommand()) {
+    case REQUEST_COMMAND: {
+    	SOAR_PRINT("HDI Recieve Request\n");
```
cjchanx: remove soar print

Components/FlightControl/FlightTask.cpp
```diff
@@ -64,13 +64,13 @@ void FlightTask::Run(void * pvParams)
 
         // Since FlightTask is so critical to managing the system, it may make sense to make this a Async task that handles commands as they come in, and have these display commands be routed over to the DisplayTask
         // or maybe HID (Human Interface Device) task that handles both updating buzzer frequencies and LED states.
-        GPIO::LED1::On();
-        GPIO::LED2::On();
-        GPIO::LED3::On();
+//        GPIO::LED1::On();
+//        GPIO::LED2::On();
+//        GPIO::LED3::On();
         osDelay(500);
```
cjchanx: Let's make sur  at least one of these is blinking every second so we know the board is running right (doesn't necessarily need to be in HDI, can be in WatchdogTask)

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+	while (1) {
+	        Command cm;
+
+	        //Wait forever for a command
+	        qEvtQueue->ReceiveWait(cm);
+
+	        //Process the command
+	        HandleCommand(cm);
+	    }
```
cjchanx: Keep in mind we might want this buzz / blink to loop constantly iterate instead of just doing this on state transitions. But this may not be needed.

Try to change it to support this structure.

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,186 @@
...
+        osDelay(2500);
+        osDelay((blinkSequence.delayMs)/blinkSequence.numBlinks);
+        osDelay(2500);
+
+
+        GPIO::LED1::Off();
+        htim2.Instance->CCR1 = 0;
+
+        osDelay(blinkSequence.delayMs);
+    }
```
AgamAulakh: Instead of calling both HAL_TIM_PWM_Start() and changing the duty cycle using the CCR1 value, let's just toggle the buzzer on and off using the HAL functions. We'll have to test this, but we should be able to just set `htim2.Instance->CCR1 = 200` at the start of the infinite while loop in Run() and then call `HAL_TIM_PWM_Start / Stop()` along with `GPIO::LED1::On() / Off()` between the delay periods. We should also add a comment explaining where the CCR1 value comes from (ie. how to calculate it using the buzzer resonant frequency).

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include "FlightTask.hpp"
+#include "Command.hpp"
+#include "etl/map.h"
+
+
+extern TIM_HandleTypeDef htim2;
+
+etl::map<RocketState, BLINK, 10> stateBlinks = etl::map<RocketState, BLINK, 10>{
+    {RS_PRELAUNCH, {2, 1000}},
+    {RS_FILL, {3, 1000}},
```
AgamAulakh: The delayMS field in the BLINK struct isn't being used like how the name implies. In BuzzBlinkSequence(), you're trying to use delayMS to put a time limit on when the numBlinks should be performed. That is, RS_PRELAUNCH will beep/flash twice in 1000ms, RS_FILL will beep/flash three times in 1000ms, etc. Though this method could be used to calculate a correct delay, the struct is misleading. It's also prone to error: consider a fake state RS_ONE mapped to BLINK {2, 1000} and RS_TWO mapped to BLINK {4, 2000} -- they would look and sound the same!

Instead, you should just define the periods you intend to pass on to osDelay. That way, we would know the exact time a buzzer/LED should be ON and OFF for.

AgamAulakh: A few more notes:

- RS_ABORT should be very distinct and easy to differentiate from the others
- RS_TEST state should be added

AgamAulakh (changes requested):

Components/FlightControl/Inc/HDITask.hpp
```diff
@@ -0,0 +1,64 @@
...
+    FILL,
+    ARM,
+    IGNITION,
+    LAUNCH,
+    BURN,
+    COAST,
+    DESCENT,
+    RECOVERY,
+    ABORT
+};
```
AgamAulakh: Enum looks like it's not being used. Let's remove it and stay consistent with RocketState in RocketSM.hpp

graceejangg: enum STATE_CHANGE is being used in the ```HandleRequestCommand(uint16_t taskCommand)``` in theHDI. cpp file. The other way to have it is the way that RocketSm.cpp has it set up where in line 110 (switch (```rs_currentState->GetStateID())```). However, this way is already implemented when we call on HDI in the RocketSm.cpp in line 77 (```    HDITask::Inst().SendCommand(Command(REQUEST_COMMAND, rs_currentState->GetStateID()));```).

AgamAulakh: RocketSM.cpp:77 ("`HDITask::Inst().SendCommand(Command(REQUEST_COMMAND, rs_currentState->GetStateID()));`") is sending the RocketState enum to HDI. HDI.cpp:119 is interpreting it as a `STATE_CHANGE` enum. To make sure these don't go out of sync, let's just use remove STATE_CHANGE and just use RocketState.

You also don't need to use the switch case when you remove STATE_CHANGE. Just assert `taskCommand` in `HDITask::HandleRequestCommand()` is within enum bounds (ie. >= RS_PRELAUNCH && < RS_NONE), and then update a local variable in HDI task to be equal to the new state received (like hdi_currentState = taskCommand). Then, BuzzBlinkSequence could just use the values mapped to hdi_currentState.

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,186 @@
...
+{
+}
+
+/**
+* @brief Initialize the HDITask
+*/
+void HDITask::InitTask()
+{
+    // Make sure the task is not already initialized
+    SOAR_ASSERT(rtTaskHandle == nullptr, "Cannot initialize flight task twice");
```
cjchanx: HDI task not flight task

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,186 @@
...
+            (uint16_t)HDI_TASK_STACK_DEPTH_WORDS,
+            (void*)this,
+            (UBaseType_t)HDI_TASK_RTOS_PRIORITY,
+            (TaskHandle_t*)&rtTaskHandle);
+
+    SOAR_ASSERT(rtValue == pdPASS, "HDITask::InitTask() - xTaskCreate() failed");
+}
+
+
+BLINK num = stateBlinks[RS_ABORT];
```
cjchanx: This variable should be a class variable, instead of floating in the cpp. 

Please rename to something more descriptive eg. `currentConfig`

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,186 @@
...
+/**
+* @brief Handles a Request Command
+* @param taskCommand The command to handle
+*/
+void HDITask::HandleRequestCommand(uint16_t taskCommand)
+{
+    //Switch for task specific command within DATA_COMMAND
+    switch (taskCommand) {
+    case PRELAUNCH:
+    	SOAR_PRINT("HDI Receive PreLaunch\n");
```
cjchanx: nit: remove this SOAR_PRINT or make it a general one, it's unusual to have it only for prelaunch

Components/FlightControl/Inc/HDITask.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include "Task.hpp"
+#include "SystemDefines.hpp"
+#include "RocketSM.hpp"
+
+ 
+
+struct BLINK{
+    uint8_t numBlinks;
+    uint16_t delayMs;
+};
```
cjchanx: I think `HDIConfig` would be a more applicable name for this struct

Components/FlightControl/RocketSM.cpp
```diff
@@ -67,6 +68,8 @@ RocketState RocketSM::TransitionState(RocketState nextState)
     // Assert the next state is initalized
     SOAR_ASSERT(rs_currentState != nullptr, "rs_currentState is nullptr in TransitionState");
 
+    HDITask::Inst().SendCommand(Command(REQUEST_COMMAND, rs_currentState->GetStateID()));
+    SOAR_PRINT("State sent to HDITask\n");
```
cjchanx: Remove this SOAR_PRINT

cjchanx (changes requested):

Components/FlightControl/FlightTask.cpp
```diff
@@ -64,6 +64,14 @@ void FlightTask::Run(void * pvParams)
...
         // Since FlightTask is so critical to managing the system, it may make sense to make this a Async task that handles commands as they come in, and have these display commands be routed over to the DisplayTask
         // or maybe HID (Human Interface Device) task that handles both updating buzzer frequencies and LED states.
+//        GPIO::LED1::On();
+//        GPIO::LED2::On();
+//        GPIO::LED3::On();
+        osDelay(500);
+//        GPIO::LED1::Off();
+//        GPIO::LED2::Off();
+//        GPIO::LED3::Off();
+        osDelay(500);
```
cjchanx: These delays and the commented out LED toggles can be removed, the 1-second LED will be handled by WatchdogTask

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,139 @@
+/**
+******************************************************************************
+* File Name          : HDITask.cpp
+* Description        : Primary flight task, default task for the system.
```
cjchanx: nit: description wrong

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+* @brief Instance Run loop for the Flight Task, runs on scheduler start as long as the task is initialized.
+* @param pvParams RTOS Passed void parameters, contains a pointer to the object instance, should not be used
+*/
+void HDITask::Run(void * pvParams)
+{
+
+	while (1) {
+		uint8_t value = 200; // the value for the duty cycle
```
cjchanx: nit: let's make `value` a constexpr at the top of this file eg.
`constexpr uint16_t BUZZER_DEFAULT_DUTY_CYCLE = 200;`

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+* @brief Handles a Request Command
+* @param taskCommand The command to handle
+*/
+void HDITask::HandleRequestCommand(uint16_t taskCommand)
+{
+    //Switch for task specific command within DATA_COMMAND
+	currentConfig = stateBlinks[RS_ABORT];
```
cjchanx: This line can be removed

Components/FlightControl/HDITask.cpp
```diff
@@ -0,0 +1,139 @@
...
+	if((RocketState)taskCommand >= RS_PRELAUNCH && (RocketState)taskCommand< RS_NONE){
+		BuzzBlinkSequence(stateBlinks[(RocketState)taskCommand]);
+		currentConfig = stateBlinks[(RocketState)taskCommand];
```
cjchanx: nit: can be refactored to be a tiny bit simpler / more efficient:
```
currentConfig = stateBlinks[(RocketState)taskCommand];
BuzzBlinkSequence(currentConfig);
```

Components/FlightControl/HDITask.cpp
graceejangg: LED is being interferred with Watchdog and FlightTask, but it has been tested that it works. The buzzer sequence is correct and will give a certain amount of beeps then a total of 8 seconds of silence before the next cycle of beeps is turned on. Tested that it reaches different states with buzzer and LED.

cjchanx (changes requested):

cjchanx (approved):

Jgerbrandt (approved):

PR #78, opened 2023-06-21
Andromeda v4.5.0 (Flash Storage, Logging, State Recovery, State Machine Updates)

cjchanx:


Components/FlightControl/FlightTask.cpp
```diff
@@ -43,8 +46,26 @@ void FlightTask::InitTask()
...
+    SPIFlash::Inst().Init();
+
+    //Get the latest state from the system storage
+    SystemState sysState;
+    bool stateReadSuccess = SystemStorage::Inst().Read(sysState);
+
+    if (stateReadSuccess == false) {
+        // Succeded to read state, initialize the rocket state machine
+		//TODO: Check if we really want to enter the state or not, or if we need to make it selective based on the state
+        rsm_ = new RocketSM(sysState.rocketState, true);
```
cjchanx, line 59: Whether to re-enter the state is a tough one to figure out: we'll need to test what happens in various power-off stages (eg. full power cut, power brownout, fault, assert, etc.) and make sure it works for every state.

We can also optionally store if the state has been successfully entered to the flash storage, but even if its been entered it may or may not need to enter it again depending on how the board reset

Components/FlightControl/RocketSM.cpp
```diff
@@ -180,6 +189,15 @@ RocketState PreLaunch::OnEnter()
...
 {
     // We don't do anything upon exiting prelaunch
+
+	//TODO: Test code for testing hardfault, make sure commented out / removed after tests
+    //uint8_t* crash = NULL;
+    //*crash = 5;
+
+    //if(&crash) {
+        //return RS_ABORT;
+    //}
```
cjchanx, line 204: Remember to remove after testing

Components/Flash/Inc/FlashTask.hpp
```diff
@@ -0,0 +1,68 @@
...
+#ifndef SOAR_FLASHTASK_HPP_
+#define SOAR_FLASHTASK_HPP_
+#include "Task.hpp"
+#include "SystemDefines.hpp"
+#include "SystemStorage.hpp"
+#include "SPIFlash.hpp"
+
+/* Macros/Enums ------------------------------------------------------------*/
+constexpr uint16_t MAX_FLASH_TASK_WAIT_TIME_MS = 5000; // The max time to wait for a command before maintenance is checked
+constexpr uint8_t FLASH_OFFSET_WRITES_UPDATE_THRESHOLD = 5; // The number of writes to flash before offsets are updated in flash
```
cjchanx, line 16: Note: Remember to use Protocol encoding or something similar (maybe just the raw data frame wrapped in COBS) to be able to recover after unaligned offset writes happen

We should add flash maintenance commands to control messages (flash format etc.)

We could also do a selectiveWrite-verify (read at the first 0xFF section) for the first ever write.

Best option is probably to make a SpannedStorage class that handles any logic for this

Components/Flash/FlashTask.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * @brief Instance Run loop for the Flash Task, runs on scheduler start as long as the task is initialized.
+ * @param pvParams RTOS Passed void parameters, contains a pointer to the object instance, should not be used
+ */
+void FlashTask::Run(void * pvParams)
+{
+    // Wait until the flash has been initialized by flight task
+    while (!SPIFlash::Inst().GetInitialized())
+        osDelay(1);
```
Matt4890, line 54: Use braces >:(

Components/FlightControl/FlightTask.cpp
Matt4890: Assuming RAM has reset (which it will have), I think re-entering might be the right thing to do. Definitely need to test this though.

Components/FlightControl/TimerTransitions.cpp
```diff
@@ -59,23 +61,24 @@ void TimerTransitions::RecoverySequence() {
...
+////        FlightTask::Inst().SendCommand(Command(CONTROL_ACTION, RSC_IGNITION_TO_LAUNCH));
+//        Timer::DefaultCallback(rtTimerHandle);
+//    }
+//    else {
+//    	Inst().ignitionCountdown->ResetTimer();
+//        FlightTask::Inst().SendCommand(Command(CONTROL_ACTION, RSC_GOTO_ARM));
+//    }
+//
+//    return;
+//}
```
Matt4890, line 83: Probably better to remove than comment out... Unless there's a good reason?

Components/Sensors/BarometerTask.cpp
```diff
@@ -94,7 +94,7 @@ void BarometerTask::Run(void * pvParams)
 
         //Wait forever for a command
         qEvtQueue->ReceiveWait(cm);
-
+
```
Matt4890, line 97: Whitespace >:(

Matt4890 (approved):
Bit of a skim review, but overall seems workable. Definitely needs testing and to address the 2 comments Chris left.

Components/Flash/Inc/SimpleDualSectorStorage.hpp
```diff
@@ -0,0 +1,260 @@
...
+ *
+ * @tparam T type of data to store
+ */
+template<typename T>
+SimpleDualSectorStorage<T>::SimpleDualSectorStorage(Flash* flashDriver, uint32_t startAddr) :
+    sector1_(flashDriver, startAddr),
+    sector2_(flashDriver, startAddr + flashDriver->GetSectorSize())
+{
+    // The startAddr must align with a sector boundary
+    SOAR_ASSERT(startAddr % flashDriver->GetSectorSize() == 0, "Error-SDSS: Invalid startAddr");
```
cjchanx, line 83: Possible mod by 0 😱 

Should be fixed with UartDriver PR

PR #81, opened 2023-08-09
Andromeda v4.6.2 (MEVManager)

Jgerbrandt:


Components/FlightControl/MEVManager.cpp
```diff
@@ -0,0 +1,30 @@
...
+    PBBRxProtocolTask::SendPBBCommand(Proto::PBBCommand::Command::PBB_CLOSE_MEV);
+}
+
+void MEVManager::HandleMEVTelemetry(Proto::TelemetryMessage& msg) {
+    if (shouldMevBeOpen == INDETERMINATE || !msg.has_mevstate()) {
+        // Do nothing
+        return;
+    }
+
+    Proto::MEVState mevState = msg.mevstate();
```
cjchanx: Because its likely this stack obj doesnt get optimized out, use .mevstate() directly

(Using a reference may be applicable but only if .mevstate() returns a reference as well)

Components/FlightControl/MEVManager.cpp
```diff
@@ -0,0 +1,30 @@
...
+    if (!mevState.mev_open() && shouldMevBeOpen == OPEN) {
+        OpenMEV();
+    } else if (mevState.mev_open() && shouldMevBeOpen == CLOSE) {
+        CloseMEV();
+    }
+}
+
+MEVManager::MEVState MEVManager::shouldMevBeOpen = INDETERMINATE;
```
cjchanx: nit: I usually prefer to put static variable initialization above all the functions

Matt4890:
This is amazing programming! Good job!!!

graceejangg:
Best piece of code I've seen in my life. WOWWW

cjchanx (approved):
Whoa this is crazy!

PR #98, opened 2024-12-29
Data Broker

shivamdesai04:


Components/DataBroker/Inc/DataBroker.hpp
```diff
@@ -0,0 +1,173 @@
...
+ * CLASS DEFINITIONS
+ ************************************/
+class DataBroker {
+ public:
+  /**
+   * @brief Publish data of a certain type
+   *         NOTE: You must ensure that there is a publisher for that type
+   */
+  template <typename T>
+  static void Publish(T* dataToPublish) {
```
cjchanx, line 43: Consider making a static instance of the class instead of making everything in the class static

cjchanx (changes requested):

PR #100, opened 2025-04-10
BAT_EN to high for umbilica power

shivamdesai04:


shivamdesai04 (commented):
looks fire, im kinda the goat 😫🤑🥵

